PR #1, opened 2015-01-23
Changed SetupWizard feature to use kickstart

bdunogier:
> Required by CI changes in ezsystems/ezpublish-community#223

Makes the SetupWizard scenarii expect `kickstart.ini` that was added in the ezpublish-community PR above.


andrerom:
+1 for temporary making sure travis runs continue to work.

joaoinacio:
See my comments on https://github.com/ezsystems/ezpublish-community/pull/223:

> I'd rather have an additional setupWizard suite named kickstart to use if really necessary, and leave the current setup tests as-is.

So FWIW -1 on removing sentences, an alternative approach would be to add a specific feature for kickstart setup IMHO.

PR #2, opened 2015-01-30
Get tests running

bdunogier:
> [EZP-23965](https://jira.ez.no/browse/EZP-23965)

Added dependencies, .travis.yml, phpunit.xml...

We have a hard requirement on ezpublish-kernel, but we would need to expose more interfaces to get rid of it. Not sure it's worth it.

## TODO
- [x] Get tests to pass ;-)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,36 @@
+language: php
+
+# run tests on php misc php versions
+php:
+  - 5.4
+  - 5.5
+  - 5.6
+
+# test only master (+ Pull requests)
+branches:
+  only:
+    - master
+
+# setup requirements for running unit tests
+before_script:
+  # TEMP get latests version of composer which is faster
+  ##- composer self-update
+  # Disable xdebug to speed things up as we don't currently generate coverge on travis
+  - if [ $TRAVIS_PHP_VERSION != "hhvm" ]; then phpenv config-rm xdebug.ini ; fi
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -0,0 +1,34 @@
...
+
+# test only master (+ Pull requests)
+branches:
+  only:
+    - master
+
+# setup requirements for running unit tests
+before_script:
+  # TEMP get latests version of composer which is faster
+  - composer self-update
```
andrerom: not needed anymore I think, this was the case in early december

composer.json
```diff
@@ -10,12 +10,25 @@
     ],
     "minimum-stability": "stable",
     "require": {
-        "ezsystems/ezpublish-legacy": ">=2014.11"
+        "ezsystems/ezpublish-legacy": ">=2014.11",
+        "ezsystems/ezpublish-kernel": "dev-master"
```
andrerom: should maybe be in dev below?

andrerom:
besides nitpicks, +1

composer.json
bdunogier: Yeah, i considered it... but technically, it *is* required, not only in dev (since it uses interfaces/classes from there (need to double check that, maybe only tests do)).

andrerom: ok

bdunogier:
All green ! I'll squash this a bit :-)

PR #3, opened 2015-01-31
Fixed SensioLabs Insight failures

bdunogier:
> See ezsystems/ezpublish-kernel#1152


andrerom:
didn't fully get this one, does it need to be done because of changes in the linked PR? Or is it just a followup?

Otherwise no service definition needed to be updated for this?

bdunogier:
It's just that some of the insight failures were on Legacy things, meaning one more PR with my other insight fix PR on ezpublish-kernel.

About services definitions, now that you mention it...  weird that it didn't blow up.

PR #4, opened 2015-02-03
EZP-24048: Don't clear legacy content cache on symfony cache:clear

andrerom:
Issue: https://jira.ez.no/browse/EZP-24048

This is also used on composer install/update, but it is not clearing content (http + stash) cache for platform stack, so would make sense to align handling in LegacyBridge/Bundle as well to avoid slow operation and  impact on performance. Instead asking users (like always), to selectively clear content based cache based on the changes they deploy.

#### Testing on eZ Publish 5.x (and not master, aka 6.x, aka eZ Platform)
If you are on eZ Publish 5.x, then apply this fix to [vendor/ezsystems/ezpublish-kernel/eZ/Bundle/EzPublishLegacyBundle/Cache/LegacyCachePurger.php](https://github.com/ezsystems/ezpublish-kernel/blob/v2014.11.0/eZ/Bundle/EzPublishLegacyBundle/Cache/LegacyCachePurger.php)

### DIFF
--- a/bundle/Cache/LegacyCachePurger.php
+++ b/bundle/Cache/LegacyCachePurger.php
@@ -19,8 +19,8 @@ use eZScript;
 use eZCache;
 
 /**
- * Purger for legacy cache.
- * Hooks into cache:clear command.
+ * Purger for legacy file based cache.
+ * Hooks into cache:clear command, which again is used by composer install and update.
  */
 class LegacyCachePurger implements CacheClearerInterface
 {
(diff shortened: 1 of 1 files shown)


bdunogier:
Why not indeed. Not against the principle.

andrerom:
review ping @bdunogier @yannickroger @glye @lolautruche

lolautruche:
+1

glye:
+1

bdunogier:
+1

PR #5, opened 2015-02-03
Fix EZP-23978: Inject index_page and default_page settings

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-23978

Inject `IndexPage` and `DefaultPage` if set, regardless of using multisite settings (content.tree_root.location_id) or not.

### DIFF
--- a/bundle/LegacyMapper/Configuration.php
+++ b/bundle/LegacyMapper/Configuration.php
@@ -277,10 +277,21 @@ class Configuration extends ContainerAware implements EventSubscriberInterface
     private function getMultiSiteSettings()
     {
         $rootLocationId = $this->configResolver->getParameter( 'content.tree_root.location_id' );
+        $indexPage = $this->configResolver->getParameter( 'index_page' );
         $defaultPage = $this->configResolver->getParameter( 'default_page' );
         if ( $rootLocationId === null )
         {
-            return array();
+            // return SiteSettings if there is no MultiSite (rootLocation is not defined)
+            $result = array();
+            if ( $indexPage !== null )
+            {
+                $result['site.ini/SiteSettings/IndexPage'] = $indexPage;
+            }
+            if ( $defaultPage !== null )
+            {
+                $result['site.ini/SiteSettings/DefaultPage'] = $defaultPage;
+            }
+            return $result;
         }
 
         $pathPrefix = trim( $this->urlAliasGenerator->getPathPrefixByRootLocationId( $rootLocationId ), '/' );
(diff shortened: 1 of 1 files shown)


andrerom:
why is the settings set twice?

joaoinacio:
MultiSiteSettings is dependent on content_tree.rootnode, if rootLocationId is null it will not set //any// setting.
So, `index_page` and `default_page` should be set independently.

If they are not defined, getMultiSiteSettings will use the default value of `/content/view/full/$id`.

Perhaps there is a cleaner way for this, but AFAICT the logic seems right?

bundle/LegacyMapper/Configuration.php
```diff
@@ -180,6 +180,9 @@ public function onBuildKernel( PreBuildKernelEvent $event )
         // Multisite settings (PathPrefix and co)
         $settings += $this->getMultiSiteSettings();
 
+        // SiteAccess settings
+        $settings += $this->getSiteSettings();
```
andrerom: my point was that this line will overwrite the same settings coming from getMultiSiteSettings()

joaoinacio: If there is multisite, yes it will use the same value again (this could be cleanup up) but if there is no multisite they won't be set yet.

I guess getMultiSiteSettings() should set legacy IndexPage and DefaultPage only if there is no custom setting, is that correct?

andrerom: My point is, it is to complex, it is hard to read.

maybe *remove* `getSiteSettings`, and move the logic into the `if ( $rootLocationId === null )` block, with inline doc saying we return plan site settings if rootLocationId is not used.

joaoinacio: Sounds good, thanks for the feedback!
Will update asap.

joaoinacio:
(Rebased to master, travis fail seems unrelated)

bdunogier:
The failure's mine, fixing it.​

andrerom:
+1

lolautruche:
+1

bundle/LegacyMapper/Configuration.php
```diff
@@ -277,10 +277,21 @@ private function getImageSettings()
...
     {
         $rootLocationId = $this->configResolver->getParameter( 'content.tree_root.location_id' );
+        $indexPage = $this->configResolver->getParameter( 'index_page' );
         $defaultPage = $this->configResolver->getParameter( 'default_page' );
         if ( $rootLocationId === null )
         {
-            return array();
+            // return SiteSettings if there is no MultiSite (rootLocation is not defined)
+            $result = array();
+            if ( $indexPage )
```
bdunogier: Don't we usually go for stricter tests ?

joaoinacio: The reasoning behind this as is, is that the setting should already be valid if set, otherwise there it would be an issue in the configuration parser, right?
In any case it could be changed to something like is_string, not empty?

bdunogier: I was just thinking about something such as `!== null`, if null is what we might get. We usually don't use very "fuzzy" tests like this, but yes, the value is validated.

joaoinacio: right, done!

PR #6, opened 2015-02-06
Fix EZP-23953: Legacy session events not triggered any more

lolautruche:
> https://jira.ez.no/browse/EZP-23953
> https://jira.ez.no/browse/EZP-23904

As of 5.3, authentication is completely managed through Symfony. One side effect is that legacy session events are not triggered any more (either via `ezpEvent` or `eZSession::triggerCallback()`), and thus listeners won't be executed. This is because it is now entirely up to Symfony to deal with the session, including session migration after authentication.
**Concrete illustration of this regression :** When using the legacy shop module, items in the shopping basket are lost after logging in.
In short, when the session is migrated after the user has been authenticated, the basket table, storing each entry per session ID, won't be updated at that time, resulting the loss of the basket for current user.

This PR introduces proxies for session storage and session save handler. They use legacy callbacks to trigger the appropriate events, just like what is done in legacy session handlers.

### DIFF
--- a/mvc/Session/LegacySessionProxy.php
+++ b/mvc/Session/LegacySessionProxy.php
@@ -1,0 +1,140 @@
+<?php
+/**
+ * This file is part of the eZ Publish LegacyBridge package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace eZ\Publish\Core\MVC\Legacy\Session;
+
+use Symfony\Component\HttpFoundation\Session\Storage\Proxy\AbstractProxy;
+use SessionHandlerInterface;
+use Closure;
+use ezpEvent;
+use eZDB;
+use eZSession;
+
+/**
+ * Session handler proxy for legacy.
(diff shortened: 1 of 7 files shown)


lolautruche:
Review ping @andrerom @bdunogier @yannickroger @pspanja @glye

yannickroger:
Before reading the code, just want to say thank you for the description :)

mvc/Session/LegacySessionProxy.php
```diff
@@ -0,0 +1,139 @@
...
+
+    public function open( $savePath, $sessionName )
+    {
+        $return = true;
+        if ( $this->sessionHandler )
+        {
+            $return = (bool)$this->sessionHandler->open( $savePath, $sessionName );
+        }
+
+        if ($return === true)
```
andrerom: cs

andrerom:
+1, looks great and thanks for the write up indeed :)

yannickroger:
+1

lolautruche:
This one being sensitive, I'll wait at least to have 3 *+1*

glye:
YESSSH!!!
The first commit fixes EZP-23904 for me.
Very +1.

lolautruche:
OK, just fixed the tests.

yannickroger:
@glye That is a great news !

PR #7, opened 2015-02-07
EZP-23996: moved legacy setup listener from kernel

bdunogier:
> [EZP-23966](https://jira.ez.no/browse/EZP-23996)
> Sub-task of [EZP-23966](https://jira.ez.no/browse/EZP-23994)
> Related to ezsystems/ezpublish-kernel#1180

Add the Legacy Setup Event Listener removed from ezpublish-kernel.


andrerom:
+1 ( but this should be merged first ;) )

PR #11, opened 2015-02-12
Fix EZP-23971: If a user is removed while logged in, the session for tha...

glye:
...t user will throw an exception

Catch the NotFoundException and load the anonymous user. Should this be logged?

Should we avoid hardcoding the anonymous user ID? Core\Repository\Repository does the same, ref:
https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/Repository/Repository.php#L246

https://jira.ez.no/browse/EZP-23971

### DIFF
--- a/bundle/EventListener/RequestListener.php
+++ b/bundle/EventListener/RequestListener.php
@@ -10,6 +10,7 @@
 namespace eZ\Bundle\EzPublishLegacyBundle\EventListener;
 
 use eZ\Publish\API\Repository\Repository;
+use eZ\Publish\Core\Base\Exceptions\NotFoundException;
 use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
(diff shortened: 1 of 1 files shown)


bdunogier:
Commit messages don't have to be sentences. The first criteria is that they fit in the log.

bdunogier:
But wasn't @lolautruche working on this as well ?

glye:
He was assisting, but the issue is assigned to me. (New input from 2ndline meant I could make a more sensible PR now.)

pspanja:
@glye 
> Should we avoid hardcoding the anonymous user ID?

Yes, this was kept for BC reasons, see https://github.com/ezsystems/ezpublish-kernel/pull/642. There is also an example in the code there how to get `anonymous_user_id` from config.

But (I'm not an expert here), is this really a correct way to fix this? I would think the session should be removed as well.

bundle/EventListener/RequestListener.php
```diff
@@ -69,7 +70,15 @@ public function onKernelRequest( GetResponseEvent $event )
             return;
         }
 
-        $apiUser = $this->repository->getUserService()->loadUser( $request->getSession()->get( 'eZUserLoggedInID' ) );
+        try
+        {
+            $apiUser = $this->repository->getUserService()->loadUser($request->getSession()->get('eZUserLoggedInID'));
```
pspanja: CS: space after and before parentheses.

glye:
@pspanja The configResolver stuff? OK, I can try that.

I thought loadUser() took care of the session, but really have no idea.

pspanja:
@glye Yup. No, we don't even care about session at that level.

lolautruche:
IMHO, this fixes the symptom, not the cause as @pspanja pointed out. https://github.com/ezsystems/ezpublish-kernel/pull/1185 was attempting to *fix* the cause in a much better place. The user should indeed be kicked out, and you can do that by resetting the security token.

I can try something tomorrow if you want.

glye:
@lolautruche You're saying #1185 is actually *better*? Wow :)
Yes please.

lolautruche:
I'm not saying it's better, just that you were closer to the root cause :wink:

PR #12, opened 2015-02-13
Fix EZP-23971: If a user is removed while logged in, session for that user will throw an exception

lolautruche:
> https://jira.ez.no/browse/EZP-23971

Relates to https://github.com/ezsystems/ezpublish-kernel/pull/1188

When in legacy mode, authentication is completely delegated to the legacy kernel. However currently logged in user is also injected in the Repository to make it usable with appropriate permissions.

Problem is that when the user is removed while a session for this user is still open, Symfony stack still tries to inject it in the repository, failing with a nasty exception.
This PR fixes the issue.

### DIFF
--- a/bundle/EventListener/RequestListener.php
+++ b/bundle/EventListener/RequestListener.php
@@ -9,6 +9,7 @@
 
 namespace eZ\Bundle\EzPublishLegacyBundle\EventListener;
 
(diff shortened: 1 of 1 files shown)


andrerom:
+1

glye:
+1 (Seems to me I wasn't all that far off, but devils, details, etc.)

PR #13, opened 2015-02-18
Fix EZP-24017: Security token issue using legacy bridge

lolautruche:
> https://jira.ez.no/browse/EZP-24017

Addresses the *Cannot set this token to trusted after instantiation* exception when trying to authenticate both in frontend and backend.
This is a regression caused by fix for [EZP-23953](https://jira.ez.no/browse/EZP-23953). Reason is that `LegacySessionStorage` does not inherit from `NativeSessionStorage`, and thus [session name and other session options from siteaccess configuration are not passed to it any
more](|https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/EventListener/SessionSetDynamicNameListener.php#L71-L89).

This patch fixes the issue by making `LegacySessionStorage` inherit from `NativeSessionStorage`. This is only a workaround to strictly respect its full interface, as `NativeSessionStorage` has public methods that are not part of an interface. Those public methods have been reimplemented and will act as proxies if `innerSessionStorage` is an actual instance of `NativeSessionStorage`.

Yes I know, it's a bit complicated :octocat:.

### DIFF
--- a/mvc/Session/LegacySessionStorage.php
+++ b/mvc/Session/LegacySessionStorage.php
@@ -10,6 +10,8 @@
 namespace eZ\Publish\Core\MVC\Legacy\Session;
 
 use Symfony\Component\HttpFoundation\Session\SessionBagInterface;
+use Symfony\Component\HttpFoundation\Session\Storage\MetadataBag;
+use Symfony\Component\HttpFoundation\Session\Storage\NativeSessionStorage;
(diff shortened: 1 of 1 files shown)


lolautruche:
Review ping @andrerom @bdunogier @yannickroger @pspanja

yannickroger:
+1

pspanja:
+1

PR #14, opened 2015-02-19
Fix EZP-23928: Preview cache not being updated

lolautruche:
> https://jira.ez.no/browse/EZP-23928

See https://github.com/ezsystems/ezpublish-legacy/pull/1143


andrerom:
+1

yannickroger:
+1

PR #16, opened 2015-03-10
EZP-23921 : ESI & host_uri matching

tharkun:
The LegacyMapper\Siteaccess does not handle host_uri matching. It means that the context is lost if you're using ESI.

This is why the ezurl operator does not work properly.

https://jira.ez.no/browse/EZP-23921

### DIFF
--- a/bundle/LegacyMapper/SiteAccess.php
+++ b/bundle/LegacyMapper/SiteAccess.php
@@ -11,6 +11,7 @@ namespace eZ\Bundle\EzPublishLegacyBundle\LegacyMapper;
 
 use eZ\Publish\Core\MVC\Legacy\LegacyEvents;
 use eZ\Publish\Core\MVC\Legacy\Event\PreBuildKernelWebHandlerEvent;
+use eZ\Publish\Core\MVC\Symfony\SiteAccess\Matcher\CompoundInterface;
 use eZSiteAccess;
 use Symfony\Component\DependencyInjection\ContainerAware;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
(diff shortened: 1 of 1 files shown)


andrerom:
+1, ping @lolautruche

bdunogier:
Do we have tests of this area ? If yes, it'd be cool to make sure this is use-case is covered.

tharkun:
I haven't written any test. My first idea with this PR was to have your feedback (in case i did not see any strange behavior with this fix)

bundle/LegacyMapper/SiteAccess.php
```diff
@@ -90,6 +90,24 @@ public function onBuildKernelWebHandler( PreBuildKernelWebHandlerEvent $event )
             $uriPart = array_diff( $aPathinfo, $aSemanticPathinfo, $this->getFragmentPathItems() );
         }
 
+        if ($siteAccess->matchingType == 'compound:logicalAnd(Map\Host, Map\URI)')
```
lolautruche: You're hardcoding the matching type check here. What if it was in the reverse order? Or what if `URIElement` was used instead of `Map\URI`?

bundle/LegacyMapper/SiteAccess.php
```diff
@@ -90,6 +90,24 @@ public function onBuildKernelWebHandler( PreBuildKernelWebHandlerEvent $event )
...
+        {
+            $subMatchers = $siteAccess->matcher->getSubMatchers();
+            if ($subMatchers)
+            {
+                foreach ($subMatchers as $key => $subMatcher)
+                {
+                    if ($key == 'Map\URI')
```
lolautruche: Why are you re-checking it here? You could directly access to the right matchers.

Furthermore, if you don't have any sub matchers, you should raise an exception, not only apply the following code if `$subMatchers` isn't empty.

```php
$subMatchers = $siteAccess->matcher->getSubMatchers();
if ( !$subMatchers )
{
    // Throw an exception as the situation is not normal at all
}

// Continue normally
```

bundle/LegacyMapper/SiteAccess.php
lolautruche: And it's always better to check the matcher instance here:

```php
if ( $siteAccess->matcher instanceof CompoundInterface )
{
   // ...
}

lolautruche:
Apart from the remarks, there are a few CS issues (mainly missing WS between parenthesis). Is @ezrobot dead again?

andrerom:
> Is @ezrobot dead again?

Not been setup on this repo yet.

bundle/LegacyMapper/SiteAccess.php
tharkun: Well you're right about that one.

tharkun:
@lolautruche CS issues should be fixed now.

I can't see the use case you were talking about using Map\Host and URIElement.

What do you think about the PR now ?

andrerom:
re ping @lolautruche

lolautruche:
+1, but we need a test

lolautruche:
On the other hand, there is currently no test for this mapper.

tharkun:
@lolautruche you're right there is no test to avoid breaking code in the future.

I can assure you it works fine on http://www.ceva.com and on every website of this company :)

andrerom:
+1

lolautruche:
Thanks @tharkun

PR #17, opened 2015-03-13
Fix EZP-23710: ezpublish:legacy:assets_install --relative fails on windows

glye:
Fixed by changing current working directory to the base of the relative directory before doing Finder::create()

https://jira.ez.no/browse/EZP-23710

### DIFF
--- a/bundle/Command/LegacyWrapperInstallCommand.php
+++ b/bundle/Command/LegacyWrapperInstallCommand.php
@@ -85,7 +85,10 @@ EOT
             {
                 $filesystem->mkdir( $targetDir, 0777 );
                 // We use a custom iterator to ignore VCS files
+                $currentDir = getcwd();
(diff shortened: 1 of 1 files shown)


andrerom:
(+1) but is this how it is done in symfony now as well for similar commands? I think one of them regarding asset / assetic is used as the blueprint for this one. ping @bdunogier @lolautruche

glye:
You mean this? https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Command/AssetsInstallCommand.php#L148
They do it the same way, except they got their directories right so no need to chdir :)

(Though I'm sure it can be done more elegantly than this PR.)

andrerom:
ok, well I'm +1 we do it this way in legacy fallback also.

bdunogier:
+1

PR #19, opened 2015-03-23
Fix EZP-24135: Output buffer gets cleared after legacy call

glye:
Moved hasInstance() method from \eZ\Publish\Core\MVC\Legacy\Kernel to ezpKernel
ref https://github.com/ezsystems/ezpublish-legacy/pull/1150

https://jira.ez.no/browse/EZP-24135

### DIFF
--- a/mvc/Kernel.php
+++ b/mvc/Kernel.php
@@ -68,16 +68,6 @@ class Kernel extends ezpKernel
         $this->setUseExceptions( true );
     }
 
-    /**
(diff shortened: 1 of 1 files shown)


lolautruche:
+1

andrerom:
+1 (including backports to 5.4 and 5.3)

glye:
@andrerom Backports not forgotten ;)

PR #22, opened 2015-03-30
Fix EZP-24053: Preview locks content object, and cause version edit conflict

lolautruche:
> https://jira.ez.no/browse/EZP-24053
> https://jira.ez.no/browse/EZP-24052

LegacyBridge part. See also https://github.com/ezsystems/ezpublish-kernel/pull/1216 .
This PR ensures that legacy website toolbar is loaded with appropriate template when in preview.

### DIFF
--- a/bundle/Controller/WebsiteToolbarController.php
+++ b/bundle/Controller/WebsiteToolbarController.php
@@ -10,6 +10,7 @@ namespace eZ\Bundle\EzPublishLegacyBundle\Controller;
 
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\Core\Helper\ContentPreviewHelper;
 use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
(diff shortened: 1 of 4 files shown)


lolautruche:
Review ping @andrerom @bdunogier @dpobel @yannickroger @pspanja @glye

bundle/Controller/WebsiteToolbarController.php
```diff
@@ -67,10 +76,11 @@ public function websiteToolbarAction( $locationId, Request $request )
             return $response;
         }
 
+        $content = $this->loadContentByLocationId($locationId);
```
andrerom: doesn't seem to be reused below, temporary change?

andrerom:
+1

bundle/Controller/WebsiteToolbarController.php
lolautruche: You're right. I changed my mind and forgot this.

yannickroger:
+1

PR #24, opened 2015-05-08
Prepare for new v0.2.1 release for 15.03

andrerom:
@emodric Looks ok? /cc @bdunogier 

:hand: __Do not merge, only for tagging__ :facepunch:

Closes #15

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     "minimum-stability": "stable",
     "require": {
         "ezsystems/ezpublish-legacy": ">=2014.11",
-        "ezsystems/ezpublish-kernel": "dev-master"
+        "ezsystems/ezpublish-kernel": "~6.0.0@alpha"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


emodric:
Everything installed correctly! :+1:

Had to specifically require `ezsystems/ezfind-ls` though...

andrerom:
> Had to specifically require ezsystems/ezfind-ls though...

I think that is to be expected now (also for eztags btw), we slimmed down requirements in ezpublish-legacy some time ago as it was pulling in everything and the kitchen sink ;)

emodric:
> I think that is to be expected now (also for eztags btw), we slimmed down requirements in ezpublish-legacy some time ago as it was pulling in everything and the kitchen sink ;)

Okay then :)

andrerom:
Release created: https://github.com/ezsystems/LegacyBridge/releases/tag/v0.2.1

PR #25, opened 2015-05-13
EZP-24278: integrated Legacy REST API in Kernel

bdunogier:
> Fixes http://jira.ez.no/browse/EZP-24278
> Requires ezsystems/ezpublish-legacy#1168

This PR adds the required elements (kernel loader, route, controller...) to run the V1 REST API through the new stack.

It overrides the ezcResponseWriter used by MVC tools in the kernel loader in order to go build a proper Response out of it. The usage of a static method is arguable...

## TODO
- [ ] Rewrite rules in ezpublish-community + ezplatform (legacy-bridge) to let REST V1 requests go through the regular index.php

### DIFF
--- a/bundle/Controller/LegacyRestController.php
+++ b/bundle/Controller/LegacyRestController.php
@@ -1,0 +1,51 @@
+<?php
+/**
+ * File containing the LegacyRestController class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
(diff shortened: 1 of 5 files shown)


andrerom:
I guess ezplatform rewrite rules can be adjusted as well for legacy bridge use, afaik it still contains legacy parts for the time being :)

bdunogier:
> I guess ezplatform rewrite rules can be adjusted as well for legacy bridge use, afaik it still contains legacy parts for the time being :)

Indeed.

lolautruche:
+1

andrerom:
+1

PR #26, opened 2015-05-21
Fix EZP-24367: Token error with remember me activated in Legacy and Symfony

lolautruche:
> https://jira.ez.no/browse/EZP-24367

RememberMeToken doesn't accept to be manually set authenticated, which is done by legacy mode. This is not needed as all authentication/authorization process is fully done by legacy in that case.

This PR removes the `setAuthenticated()` call on the token. Note that in the profiler toolbar, username will appear in yellow (not authenticated). This is perfectly safe as security is fully handled by legacy kernel.
Further more, this PR also ensures that RememberMeListener is never called when in legacy mode.

### DIFF
--- a/bundle/DependencyInjection/Compiler/RememberMeListenerPass.php
+++ b/bundle/DependencyInjection/Compiler/RememberMeListenerPass.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * This file is part of the eZ LegacyBridge package.
+ *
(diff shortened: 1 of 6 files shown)


lolautruche:
Review ping @bdunogier @andrerom @dpobel @yannickroger @pspanja

bdunogier:
+1

yannickroger:
+1

PR #27, opened 2015-06-08
Fix EZP-24457: Change Request injection to RequestStack

lolautruche:
> https://jira.ez.no/browse/EZP-24457

See https://github.com/ezsystems/ezpublish-kernel/pull/1286


lolautruche:
Review ping @andrerom @bdunogier @pspanja @yannickroger @glye

pspanja:
+1 when it passes :)

andrerom:
+1 when travis is ok on it.

PR #28, opened 2015-06-09
Fix EZP-24463: Preview will fail with remember me activated in Legacy and Symfony

lolautruche:
> https://jira.ez.no/browse/EZP-24463

See [EZP-24367](https://jira.ez.no/browse/EZP-24367) and https://github.com/ezsystems/LegacyBridge/pull/26

Previous PR removed `$token->setAuthenticated( true )` as this is not permitted with some tokens like RememberMeToken (hence the exception before).
However, not being authenticated has a side effect. The AuthorizationChecker will indeed ensure by default that current token is authenticated before checking authorization. And as preview checks if
current user is granted permission in the PreviewController, permission is denied and the system tries to redirect to the login page.

This PR introduces `LegacyToken` which decorates current token. It will always say that it's authenticated. Note that there is no security risk as authentication has already been made by the legacy kernel.


lolautruche:
Review ping @andrerom @bdunogier @yannickroger @pspanja @glye

yannickroger:
+1

glye:
+1

PR #32, opened 2015-08-27
Switched to PSR-2

lolautruche:
+ added config file for PHP-CS-Fixer

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,29 @@
+<?php
+
+return Symfony\CS\Config\Config::create()
+    ->setUsingLinter(false)
+    ->setUsingCache(true)
+    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
+    ->fixers([
+        'concat_with_spaces',
+        '-concat_without_spaces',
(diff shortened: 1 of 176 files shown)


andrerom:
Seems ezrobot is missing fore this repo, add and force push to check?

lolautruche:
It's not configured on this repo

.php_cs
```diff
@@ -0,0 +1,28 @@
...
+        '-concat_without_spaces',
+        '-empty_return',
+        '-phpdoc_params',
+        '-phpdoc_separation',
+        '-phpdoc_to_comment',
+        '-spaces_cast',
+        '-blankline_after_open_tag',
+        '-single_blank_line_before_namespace',
+        // psr0 has weird issues with our PSR-4 layout, so deactivating it.
+        '-psr0',
```
michalpipa: I usually enable this check on Studio bundles. I didn't have any problems so far.

lolautruche: It's because of split with `lib/` and `mvc`. If you don't have that, then no problem

andrerom:
+1

PR #31, opened 2015-08-25
Fix EZP-24691: Move scope change to event so legacy-bridge can build …

wizhippo:
…the correct cli kernel

Closes #30

PR #33, opened 2015-08-28
Fix restoring working directory when building legacy CLI kernel handler

emodric:
After merging #31, and associated PR in `ezsystems/ezpublish-kernel`, I get an error when running `assets:install` command:

```
eddie@abyss: /var/www/html/netgensite $ php ezpublish/console assets:install --symlink --relative -v
                                              
  [InvalidArgumentException]                  
  The target directory "web" does not exist.  
                                              
Exception trace:
 () at /var/www/html/netgensite/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Command/AssetsInstallCommand.php:74
 Symfony\Bundle\FrameworkBundle\Command\AssetsInstallCommand->execute() at /var/www/html/netgensite/vendor/symfony/symfony/src/Symfony/Component/Console/Command/Command.php:259
 Symfony\Component\Console\Command\Command->run() at /var/www/html/netgensite/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:886
 Symfony\Component\Console\Application->doRunCommand() at /var/www/html/netgensite/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:195
 Symfony\Component\Console\Application->doRun() at /var/www/html/netgensite/vendor/symfony/symfony/src/Symfony/Bundle/FrameworkBundle/Console/Application.php:96
 Symfony\Bundle\FrameworkBundle\Console\Application->doRun() at /var/www/html/netgensite/vendor/symfony/symfony/src/Symfony/Component/Console/Application.php:126
 Symfony\Component\Console\Application->run() at /var/www/html/netgensite/ezpublish/console:27


assets:install [--symlink] [--relative] [--] [<target>]
```

This fixes the issue by making sure that the previously used working directory is restored after building CLI kernel handler, instead of restoring to web root (`web/`).

### DIFF
--- a/mvc/Kernel/Loader.php
+++ b/mvc/Kernel/Loader.php
@@ -215,6 +215,7 @@ class Loader extends ContainerAware
         {
             if ( !$that->getCLIHandler() )
             {
+                $currentDir = getcwd();
                 chdir( $legacyRootDir );
 
                 $legacyParameters = new ParameterBag( $container->getParameter( 'ezpublish_legacy.kernel_handler.cli.options' ) );
@@ -226,7 +227,8 @@ class Loader extends ContainerAware
                 $that->setCLIHandler(
                     new CLIHandler( $legacyParameters->all(), $container->get( 'ezpublish.siteaccess' ), $container )
                 );
-                chdir( $webrootDir );
+
+                chdir( $currentDir );
             }
 
             return $that->getCLIHandler();
--- a/mvc/Kernel/Loader.php
+++ b/mvc/Kernel/Loader.php
@@ -206,12 +206,11 @@ class Loader extends ContainerAware
     public function buildLegacyKernelHandlerCLI()
     {
         $legacyRootDir = $this->legacyRootDir;
-        $webrootDir = $this->webrootDir;
         $eventDispatcher = $this->eventDispatcher;
         $container = $this->container;
         $that = $this;
 
-        return function () use ( $legacyRootDir, $webrootDir, $container, $eventDispatcher, $that )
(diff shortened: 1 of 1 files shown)


andrerom:
This seems to be incomplete. webrootDir is still used several places.

emodric:
It is, but I think that is intended, right?

wizhippo:
Has this always been the case?  What does https://github.com/ezsystems/LegacyBridge/issues/31 do that changes this behavior? Why does the legacy kernel loader not restore the working director in the first place?

emodric:
I can't answer that unfortunately. Maybe your fix just revealed this issue now.

wizhippo:
@andrerom Do you mean in just the loader?  For example the same fix should be applied to building the web legacy kernel too.  webrootDir seems to only be need as a fallback for Kernel::leaveLegacyRootDir, which I'm not too sure about.  When would Kernel not have a previous previousRunningDir to switch back to as once he kernel is constructed it would be set by the calling of enterLegacyRootDir?

emodric:
I think that restoring to webroot when building web handler is intended, because, after all, working directory for the web itself IS webroot and not project root.

But ofcourse, I might be wrong.

Maybe @lolautruche can give some insight

emodric:
@andrerom @wizhippo Or maybe he meant that it was still transfered to closure but not actually used. That was an oversight.

andrerom:
I think it looks ok now, as you said in web and rest context web folder is expected when not in legacy folder.

lolautruche:
+1

PR #35, opened 2015-09-02
EZP-24691: Fixed regression when resetting the kernel and it is not yet installed/configured

lolautruche:
Fixes regression introduced in #31 
When you have a clean install (wizard not run yet), a PDO exception is thrown because the legacy kernel loader tries to load the kernel to reset it.

With this patch the kernel will be reset only if it is already initialized.

### DIFF
--- a/mvc/Kernel/Loader.php
+++ b/mvc/Kernel/Loader.php
@@ -313,12 +313,15 @@ class Loader extends ContainerAware
      */
     public function resetKernel()
     {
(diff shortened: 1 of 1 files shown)


lolautruche:
Review ping @andrerom @bdunogier @yannickroger @glye @wizhippo @emodric

glye:
+1

emodric:
Looks good! +1

andrerom:
+1

PR #36, opened 2015-09-03
Fix EZP-24761: PHP fatal error redeclaring eZUpdateDebugSettings

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24761

##### Problem:
legacy eZScript is being instantiated before kernel, which causes a redeclaration of `eZUpdateDebugSettings()`
( see https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/classes/ezscript.php#L1223 )

##### Solution:
As ezpKernelWeb already requires global_functions manually, apply the same solution here 
(ref: https://github.com/ezsystems/ezpublish-legacy/blob/master/kernel/private/classes/ezpkernelweb.php#L145 )

### DIFF
--- a/mvc/Kernel/Loader.php
+++ b/mvc/Kernel/Loader.php
@@ -210,6 +210,9 @@ class Loader extends ContainerAware
                 $currentDir = getcwd();
                 chdir($legacyRootDir);
 
+                // require global functions before cli instance.
+                require_once 'kernel/private/classes/global_functions.php';
+
                 $legacyParameters = new ParameterBag($container->getParameter('ezpublish_legacy.kernel_handler.cli.options'));
                 if ($that->getBuildEventsEnabled()) {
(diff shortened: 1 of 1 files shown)


andrerom:
+0,5, ping @lolautruche

rpmsauron:
I already tested this with João Inácio on an up-to-date master and it fixes the related jira EZP-24761.
I guess no further QA is needed.

glye:
+0.5 AFAICT

lolautruche:
This issue should have been fixed with https://github.com/ezsystems/LegacyBridge/pull/31 / https://github.com/ezsystems/ezpublish-kernel/pull/1366

One don't have to modify `initializeContainer()` as a workaround any more. 

This patch looks like a very dirty hack anyway :wink:

joaoinacio:
@lolautruche: thanks, looks like composer had stable version instead of master,
requiring legacy-bridge:dev-master seems to have resolved it with mentioned PR .

Closing

PR #37, opened 2015-10-08
Fix PreviewController::getForwardRequest method signature

emodric:
https://github.com/ezsystems/ezpublish-kernel/pull/1449 changed the method signature on `PreviewController::getForwardRequest` so legacy bridge requires the same change too.


emodric:
Failing tests are not related to this change.

bdunogier:
The error is caused by a private property in a class inherited in the LegacyBridge can't be tested propertly. Will fix later :-)

PR #39, opened 2015-10-15
Fix compatibility with Twig 2.0 by not extending deprecated Twig_Loader_String

emodric:
`Twig_Loader_String` and eZ variant `LoaderString` serve the purpose of loading legacy templates and it works fine.

Since `Twig_Loader_String` is deprecated and will be removed in Twig 2.0, this simply makes sure that our loader does not extend the original one, instead implementing required methods by itself.

Since the methods are fairly simple, and `exists()` was overriden anyhow, and also considering that this is after all *legacy* bridge, i think this is the most reasonable way to get the bundle to be compatible with Twig 2.0.

Tested manually with dev-master of `twig/twig` package.

### DIFF
--- a/mvc/Templating/Tests/Twig/LoaderStringTest.php
+++ b/mvc/Templating/Tests/Twig/LoaderStringTest.php
@@ -13,6 +13,24 @@ use PHPUnit_Framework_TestCase;
 
 class LoaderStringTest extends PHPUnit_Framework_TestCase
 {
+    public function testGetSource()
+    {
+        $loaderString = new LoaderString();
+        $this->assertSame('foo', $loaderString->getSource('foo'));
+    }
+
+    public function testGetCacheKey()
+    {
+        $loaderString = new LoaderString();
+        $this->assertSame('foo', $loaderString->getCacheKey('foo'));
+    }
+
+    public function testIsFresh()
+    {
+        $loaderString = new LoaderString();
+        $this->assertSame(true, $loaderString->isFresh('foo', time()));
+    }
+
     /**
      * @dataProvider existsProvider
      */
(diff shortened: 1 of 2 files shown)


yannickroger:
ping @lolautruche

andrerom:
LGTM

lolautruche:
+1

emodric:
@yannickroger Merging this ? :)

yannickroger:
@emodric Can you check why the test are failing ?

emodric:
@yannickroger The failure is unrelated to this change. @bdunogier explained why they're failing at https://github.com/ezsystems/LegacyBridge/pull/37#issuecomment-146695817

The relevant tests, the ones I added, are passing, as well as other string loader tests :)

yannickroger:
Thanks @emodric

emodric:
Thank you @yannickroger  :)

lolautruche:
Can this be backported to 5.4 please?

andrerom:
Why? Twig 2.0 does not work on any of the officially recommended PHP versions supported by 5.4.

lolautruche:
Because 5.4 works perfectly with PHP 7 and the project I'm having the issue with is running 5.4 with PHP 7 😉 
And also because it doesn't hurt BC

bdunogier:
I'm fine with a backport to 5.4.

andrerom:
Feel free, but we need an issue for that :)

emodric:
@bdunogier FYI there are a couple of Twig 2 compatibility PRs in the repo, not just this one :)

lolautruche:
@andrerom There is already one, with customer ticket 😉 : https://jira.ez.no/browse/EZP-26966

@emodric Yeah, I already mentioned it, but it doesn't sound clear in the issue. Gonna add a comment.

PR #42, opened 2015-11-18
Update .travis.yml to try to fix memory issues on legacy package

andrerom:


andrerom:
It breaks, but because of broken Preview tests, so merging this to get proper errors on other PR's.

PR #43, opened 2015-11-18
Fix tests that error out

emodric:
Adds a missing constructor argument in test.

Depends on: https://github.com/ezsystems/ezpublish-kernel/pull/1506


bdunogier:
Hmm, Travis looks very broken on this one...

I'm just gonna trust you :-)

emodric:
It does, but it requires the patch in kernel too :)

Next PR should benefit (with only one broken test for previews :D)

PR #44, opened 2015-11-19
Adjust view provider priorities to come before the ones in kernel that provide default templates

emodric:
Now that eZ kernel always provides the default templates (https://github.com/ezsystems/ezpublish-kernel/pull/1496), this will never execute, so raising priorities to make it work again.


bdunogier:
Looks good.

emodric:
Thanks! \o/

bdunogier:
Thank you :-)

PR #45, opened 2015-11-19
Remove Symfony 2.8 deprecation notices

emodric:
This removes three Symfony 2.8 deprecation notices:

1) Notice about extending deprecated `ContainerAwareInterface`
2) Notice about using `scope` parameter in service definitions
3) Notice about using deprecated non quoted service references that start with `@`


andrerom:
+0,5 ping @bdunogier @lolautruche

lolautruche:
+1

emodric:
And the tests finally pass for legacy bridge :dancers: :metal:

andrerom:
Thanks @emodric :)

emodric:
Thank you :)

PR #49, opened 2015-12-17
Adjust block view provider priority

emodric:
This PR adjusts block view provider priority to come before the one in kernel that provides default templates.

Now that eZ kernel always provides the default templates (ezsystems/ezpublish-kernel#1496), this will never execute, so raising priority to make it work again.

This PR is continuation of https://github.com/ezsystems/LegacyBridge/pull/44. It should've been included there.

### DIFF
--- a/bundle/Resources/config/view.yml
+++ b/bundle/Resources/config/view.yml
@@ -37,7 +37,7 @@ services:
         calls:
             - [setPageService, ["@ezpublish.fieldType.ezpage.pageService"]]
         tags:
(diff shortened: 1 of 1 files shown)


emodric:
Ping @bdunogier :)

emodric:
Bloody Composer! :)

FWIW, tests passed on my machine :)

bdunogier:
Really thinking that we might want to actually diminish the default ones... what do you think ?

emodric:
Well, default ones are at -1. They could go to -255. They are, after all, fallback.

PR #52, opened 2016-03-01
[Travis] Disable memory limit on composer install

andrerom:


emodric:
+1 :)

bdunogier:
+1

andrerom:
well.. it fails and I havn't had time to look into why, so you're +1's are not real +1's they are just to try to get me to finish this :) Input on how this needs to be is welcome )

emodric:
Why not just do `curl -sS https://getcomposer.org/installer | php` and then `php -d memory_limit=-1 composer.phar install` ?

andrerom:
could do that, but on the other hand it is passing now :)

emodric:
Great! :)

PR #53, opened 2016-04-14
Add support for php7 and get all tests running

andrerom:
~~Might fail on a few things.~~ *passing*

*Note: PHP7 might never be fully supported with legacy, this just makes sure it can be used for dev use, and is a first step towards getting legacy running on php7.*

### DIFF
--- a/bundle/Cache/SwitchableHttpCachePurger.php
+++ b/bundle/Cache/SwitchableHttpCachePurger.php
@@ -24,13 +24,13 @@ class SwitchableHttpCachePurger implements GatewayCachePurger
         $this->gatewayCachePurger = $gatewayCachePurger;
     }
 
-    public function purge($cacheElements)
+    public function purge($locationIds)
     {
         if ($this->isSwitchedOff()) {
-            return $cacheElements;
+            return $locationIds;
         }
 
-        return $this->gatewayCachePurger->purge($cacheElements);
+        return $this->gatewayCachePurger->purge($locationIds);
     }
 
     public function purgeAll()
@@ -42,12 +42,12 @@ class SwitchableHttpCachePurger implements GatewayCachePurger
         $this->gatewayCachePurger->purgeAll();
     }
 
-    public function purgeForContent($contentId)
+    public function purgeForContent($contentId, $locationIds = array())
     {
         if ($this->isSwitchedOff()) {
             return;
         }
 
-        $this->gatewayCachePurger->purgeForContent($contentId);
+        $this->gatewayCachePurger->purgeForContent($contentId, $locationIds);
     }
 }
--- a/bundle/DependencyInjection/EzPublishLegacyExtension.php
+++ b/bundle/DependencyInjection/EzPublishLegacyExtension.php
@@ -59,7 +59,7 @@ class EzPublishLegacyExtension extends Extension
         $processor = new ConfigurationProcessor($container, 'ezpublish_legacy');
         $processor->mapConfig(
             $config,
-            function (array &$scopeSettings, $currentScope, ContextualizerInterface $contextualizer) {
+            function (array $scopeSettings, $currentScope, ContextualizerInterface $contextualizer) {
                 if (isset($scopeSettings['templating']['view_layout'])) {
                     $contextualizer->setContextualParameter('view_default_layout', $currentScope, $scopeSettings['templating']['view_layout']);
                 }
--- a/composer.json
+++ b/composer.json
@@ -15,10 +15,10 @@
         "twig/twig": "^1.23 | ^2.0"
     },
     "require-dev": {
-        "matthiasnoback/symfony-dependency-injection-test": "0.*",
-        "phpunit/phpunit": "~4.1.3",
-        "mikey179/vfsStream": "1.1.0",
-        "mockery/mockery": "dev-master"
+        "matthiasnoback/symfony-dependency-injection-test": "^1.0",
+        "phpunit/phpunit": "~4.7",
+        "mikey179/vfsStream": "~1.1.0",
+        "mockery/mockery": "^0.9"
     },
     "autoload": {
         "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -12,6 +12,7 @@
     "require": {
         "ezsystems/ezpublish-legacy": ">=2014.11",
         "ezsystems/ezpublish-kernel": "~6.0@dev",
+        "sensio/distribution-bundle": "^3.0|^4.0",
         "twig/twig": "^1.23 | ^2.0"
     },
     "require-dev": {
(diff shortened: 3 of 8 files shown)


bundle/DependencyInjection/EzPublishLegacyExtension.php
```diff
@@ -59,7 +59,7 @@ public function load(array $configs, ContainerBuilder $container)
         $processor = new ConfigurationProcessor($container, 'ezpublish_legacy');
         $processor->mapConfig(
             $config,
-            function (array &$scopeSettings, $currentScope, ContextualizerInterface $contextualizer) {
+            function (array $scopeSettings, $currentScope, ContextualizerInterface $contextualizer) {
```
emodric: Is this something that could break on PHP 5.x?

andrerom: Don't see how, the variable is not modified inside the closure.

emodric: Yep, just wanted to confirm.

andrerom:
review ping @emodric @lolautruche

bundle/DependencyInjection/EzPublishLegacyExtension.php
lolautruche: No, it won't break.

bundle/LegacyMapper/SiteAccess.php
```diff
@@ -47,6 +47,10 @@ public function onBuildKernelWebHandler(PreBuildKernelWebHandlerEvent $event)
         $request = $event->getRequest();
         $uriPart = array();
 
+        // @todo Remove when, if ever, legacy does not use deprecated constructor names
+        // triggers initial autoload of eZSiteAccess and sileces deprecation warning
+        @eZSiteAccess::TYPE_DEFAULT;
```
lolautruche: I don't get this...
As for the comment, you just need to silent `E_DEPRECATED` errors in php.ini

mvc/Tests/Security/SSOListenerTest.php
```diff
@@ -83,7 +83,7 @@ public function testGetPreAuthenticatedData()
 
         $userId = 123;
         $passwordHash = md5('password');
-        $legacyUser = new eZUser(array('contentobject_id' => $userId));
+        $legacyUser = @new eZUser(array('contentobject_id' => $userId));
```
lolautruche: Same comment as above. Better manipulate php.ini to silent `E_DEPRECATED` before start.

bundle/LegacyMapper/SiteAccess.php
andrerom: We can, but for symfony 3.0 support we actually want to have deprecation errors make test fails on other things, hence the selective silencing when loading legacy classes.

mvc/Tests/Security/SSOListenerTest.php
andrerom: same reply :)

bundle/LegacyMapper/SiteAccess.php
lolautruche: `E_DEPRECATED !== E_USER_DEPRECATED`

😉

andrerom: true, ok I'll see if I get that running then.

andrerom: fixed in 90d6969

mvc/Tests/Security/SSOListenerTest.php
andrerom: did not fix this as the aim is to only silence deprecation errors regarding constructor since that is the only one we most likely can't do anything about w/o introducing BC.

So left this one here but with a comment, which I guess is ok given this is test code only.
see 90d6969

composer.json
```diff
@@ -15,10 +15,10 @@
...
     },
     "require-dev": {
-        "matthiasnoback/symfony-dependency-injection-test": "0.*",
-        "phpunit/phpunit": "~4.1.3",
-        "mikey179/vfsStream": "1.1.0",
-        "mockery/mockery": "dev-master"
+        "matthiasnoback/symfony-dependency-injection-test": "^1.0",
+        "phpunit/phpunit": "~4.7",
+        "mikey179/vfsStream": "~1.1.0",
+        "mockery/mockery": "^0.9@dev"
```
emodric: Why `@dev` ?

emodric:
Aside from inline comment, +1.

bundle/Cache/SwitchableHttpCachePurger.php
```diff
@@ -42,12 +42,12 @@ public function purgeAll()
         $this->gatewayCachePurger->purgeAll();
     }
 
-    public function purgeForContent($contentId)
+    public function purgeForContent($contentId, $locationIds = array())
```
emodric: Isn't this a breaking change? If so, we should require eZ Kernel `>=6.3.2|>=6.4.2` (when 6.4.2 is released) in `composer.json`, right?

composer.json
andrerom: https://github.com/padraic/mockery/compare/0.9.5...0.9

emodric: Okay, but isn't `^0.9` enough?

bundle/Cache/SwitchableHttpCachePurger.php
andrerom: yes, _but_ adding optional argument won't break the contract. In retrospective it was the kernel that broke the contract here.

composer.json
andrerom: it can be, but then we won't pull in those fixes as it will use the tag.

emodric: It will pull 0.9.5, that's not enough?

bundle/Cache/SwitchableHttpCachePurger.php
emodric: > So besides the fact that kernel should not have done this change, I think this fix is ok as is, it will work on older kernels also.

Agreed.

composer.json
andrerom: removed it

andrerom:
looks ok to you as well @lolautruche ?

composer.json
emodric: > does not sound like you checked the link, but removed it

I did, but the link shows diff between 0.9 and 0.9.5 versions. `^0.9` will pull in 0.9.5, so I'm not sure what the issue is :)

lolautruche:
+1

Weird that it doesn't fail with PHP 5.4

andrerom:
Found it, we allow SensioDistributionBundle 5.0 now for work towards Sf 3.0 support, it supports PHP 5.5+, and it does not contain Configurator as needed by LegacyBridge.

Added it to composer.json as it is a clear requirement.

Thanks for reviews, merging :)

PR #56, opened 2016-05-02
Fix EZP-25739: utf-8 URIs break siteaccess match / ezurl()

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-25739

Legacy siteaccess matching (and ezur()) is broken when using utf-8 characters after setting up urlalias_uri.
Basically, such uri's will then be considered as part of the siteaccess, then washed, which generates 404 urls

The problem is that utf8 pathinfo is compared to urlencoded semanticPathInfo, see also https://github.com/ezsystems/ezpublish-kernel/pull/1522

### DIFF
--- a/bundle/LegacyMapper/SiteAccess.php
+++ b/bundle/LegacyMapper/SiteAccess.php
@@ -83,7 +83,7 @@ class SiteAccess implements EventSubscriberInterface
         $pathinfo = str_replace($request->attributes->get('viewParametersString'), '', $request->getPathInfo());
         $semanticPathinfo = $request->attributes->get('semanticPathinfo', $pathinfo);
         if ($pathinfo != $semanticPathinfo) {
-            $aPathinfo = explode('/', substr($pathinfo, 1));
+            $aPathinfo = explode('/', substr(rawurldecode($pathinfo), 1));
             $aSemanticPathinfo = explode('/', substr($semanticPathinfo, 1));
             $uriPart = array_diff($aPathinfo, $aSemanticPathinfo, $this->getFragmentPathItems());
         }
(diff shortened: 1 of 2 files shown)


andrerom:
test coverage? :)

joaoinacio:
there seem to be none for SiteAccess, looking into it

joaoinacio:
ping @andrerom @bdunogier

andrerom:
+1, wdyt @bdunogier ?

glye:
+1

joaoinacio:
@andrerom: is this OK for merge or do we need @bdunogier 's feedback here?

bdunogier:
Looks good to me.
We just need to be sure we are merging it against the right version, since it is a bugfix.

joaoinacio:
@bdunogier: AFAICT there's only master and features, see also related kernel one.

bdunogier:
I'll have a deeper look, I was on my phone for the previous review/reply :-)

bdunogier:
I have bumped master to 1.0.x-dev. This can just be merged into master as we haven't bumped to 1.1 (yet).

+1

emodric:
> I have bumped master to 1.0.x-dev. This can just be merged into master as we haven't bumped to 1.1 (yet).

Are you sure? Aren't you supposed to bump the master to 1.1.x? I mean, 1.0.x is used by eZ kernel 6.0, 6.1 and 6.2.

bdunogier:
Why would we bump it @emodric ? It is a backward compatible bugfix, is it not ?

emodric:
Maybe I misunderstood... Why the talk about versions then if it's a backwards compatible fix?

bdunogier:
> Why the talk about versions then if it's a backwards compatible fix?

Because unlike features, bugfixes can actually be merged to older versions. When i asked this, i was not sure if we had multiple minor versions yet or not.

As it turns out, no need for backports here :)

All clear ?

emodric:
Yep :)

PR #59, opened 2016-07-05
Fix EZP-25960: Double layout using legacy /layout/set module and LegacyKernelController

lolautruche:
> https://jira.ez.no/browse/EZP-25960

When using `/layout/set` legacy module through `LegacyKernelController` having a module layout configured, result from legacy kernel is rendered using this module layout.
However, when using `/layout/set`, you already define a layout in legacy stack (via `layout.ini`. Problem is that the result from `/layout/set` will be embedded in the configured module layout, which can give very
unexpected results.
Best example is that when you want to render XML, JSON or anything not HTML, you'll have the result embedded in some HTML layout.)

This patch adds `_ezpublishLegacyLayoutSet` route which corresponds to `/layout/set` module, and ensures that if this route is used through legacy kernel, custom layout won't be used.
This fixes the usual behavior of layout module.

### DIFF
--- a/bundle/Resources/config/routing.yml
+++ b/bundle/Resources/config/routing.yml
@@ -26,3 +26,10 @@ _ezpublishLegacyRest:
         _controller: ezpublish_legacy.rest.controller:restAction
     requirements:
         path: .*
+
+_ezpublishLegacyLayoutSet:
+    path: /layout/set/{path}
+    defaults:
+        _controller: ezpublish_legacy.controller:indexAction
+    requirements:
+        path: .+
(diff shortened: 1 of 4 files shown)


bdunogier:
+1

bundle/Resources/config/routing.yml
```diff
@@ -26,3 +26,10 @@ _ezpublishLegacyRest:
         _controller: ezpublish_legacy.rest.controller:restAction
     requirements:
         path: .*
+
+_ezpublishLegacyLayoutSet:
+    path: /layout/set/{path}
```
andrerom: does this work with uri SA?

lolautruche: > does this work with uri SA

Yes, as we override default router tu use semanticPathinfo. So this will always work.
The requirements is just to allow `/` for `{path}`

lolautruche:
@andrerom Is this OK for you ?

lolautruche:
Any feedback please ?
@andrerom @bdunogier @yannickroger

andrerom:
+1

The two others are out catching pokemons but should be back next week ;)

yannickroger:
+1

yannickroger:
Thanks @lolautruche

PR #62, opened 2016-08-30
QA: Fix EZP-26181: Content with same name as URL matcher will lead to broken links

glye:
> Fixes https://jira.ez.no/browse/EZP-26181
> Status: QA

This is caused by the use of `array_diff` to determine the siteaccess part of the URL, here:
https://github.com/ezsystems/LegacyBridge/blob/master/bundle/LegacyMapper/SiteAccess.php#L88
When the the path-without-siteaccess contains an element that equals the siteaccess, the diff is empty, and the siteaccess is not found, and subsequently removed from all URLs.

Example: The siteaccess is `Admin` and we're requesting `/Media/Images/Admin/EZP-26181`. Before this PR, this means that:
`$pathinfo = /Admin/Media/Images/Admin/EZP-26181`
`$semanticPathinfo = /Media/Images/Admin/EZP-26181`
The paths are broken into path element arrays, and diffed:
```
array_diff(
    [Admin, Media, Images, Admin, EZP-26181],
    [Media, Images, Admin, EZP-26181]
)
```
The result of this is an empty array, as all elements of the second parameter exist in the first - array_diff() doesn't care that "Admin" is in there twice. So the `$uriPart` is empty, meaning the siteaccess was not found.

The fix is to start at the end/right of the paths, and chop off the bit that is equal. What's left, literally and figuratively, is the siteaccess uri part.

Example: After the PR, we have the same input strings:
`$pathinfo = /Admin/Media/Images/Admin/EZP-26181`
`$semanticPathinfo = /Media/Images/Admin/EZP-26181`
Now we lookup the position of `$semanticPathinfo` within `$pathinfo` directly using mb_strrpos (i.e. backwards). The `$uriPart` is taken to be whatever is to the left of that, i.e. `/Admin`.

NB: This fixes the bug according to steps-to-reproduce, but I'm not sure I'm dealing with fragment path items the right way.

- [x] Add test
- [x] Deal with the posibility of mb_strrpos() returning false? i.e. semantic pathinfo not found within pathinfo, which Should Not Happen(TM)

### DIFF
--- a/bundle/LegacyMapper/SiteAccess.php
+++ b/bundle/LegacyMapper/SiteAccess.php
@@ -81,25 +81,16 @@ class SiteAccess implements EventSubscriberInterface
 
         // uri_part
         $pathinfo = str_replace($request->attributes->get('viewParametersString'), '', rawurldecode($request->getPathInfo()));
-        $semanticPathinfo = $request->attributes->get('semanticPathinfo', $pathinfo);
-        if ($pathinfo != $semanticPathinfo) {
-            $aPathinfo = explode('/', substr($pathinfo, 1));
-            $aSemanticPathinfo = explode('/', substr($semanticPathinfo, 1));
-            $aFragmentPathInfo = $this->getFragmentPathItems();
-            if ($aFragmentPathInfo !== ['_fragment']) {
-                while (count($aFragmentPathInfo) > 0 &&
-                       end($aPathinfo) === end($aFragmentPathInfo)) {
-                    array_pop($aPathinfo);
-                    array_pop($aFragmentPathInfo);
-                }
-            } else {
-                while (count($aSemanticPathinfo) > 0 &&
-                       end($aPathinfo) === end($aSemanticPathinfo)) {
-                    array_pop($aPathinfo);
-                    array_pop($aSemanticPathinfo);
-                }
-            }
-            $uriPart = $aPathinfo;
+        $fragmentPathInfo = implode('/', $this->getFragmentPathItems());
+        if ($fragmentPathInfo !== '_fragment') {
+            $semanticPathinfo = $fragmentPathInfo;
+        } else {
+            $semanticPathinfo = $request->attributes->get('semanticPathinfo', $pathinfo);
+        }
+        $pos = mb_strrpos($pathinfo, $semanticPathinfo);
+        if ($pathinfo != $semanticPathinfo && $pos !== false) {
+            $uriPart = mb_substr($pathinfo, 1, $pos - 1);
+            $uriPart = explode('/', $uriPart);
         }
 
         // Handle host_uri match
--- a/bundle/LegacyMapper/SiteAccess.php
+++ b/bundle/LegacyMapper/SiteAccess.php
@@ -85,7 +85,21 @@ class SiteAccess implements EventSubscriberInterface
         if ($pathinfo != $semanticPathinfo) {
             $aPathinfo = explode('/', substr($pathinfo, 1));
             $aSemanticPathinfo = explode('/', substr($semanticPathinfo, 1));
-            $uriPart = array_diff($aPathinfo, $aSemanticPathinfo, $this->getFragmentPathItems());
+            $aFragmentPathInfo = $this->getFragmentPathItems();
+            if ($aFragmentPathInfo !== ['_fragment']) {
+                while (count($aFragmentPathInfo) > 0 &&
+                       end($aPathinfo) === end($aFragmentPathInfo)) {
+                    array_pop($aPathinfo);
+                    array_pop($aFragmentPathInfo);
+                }
+            } else {
+                while (count($aSemanticPathinfo) > 0 &&
+                       end($aPathinfo) === end($aSemanticPathinfo)) {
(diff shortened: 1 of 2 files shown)


glye:
Review ping @yannickroger @bdunogier @andrerom

glye:
Umlauts are not required to reproduce this. Updated issue, PR and commit to reflect that.

andrerom:
+0.5, Tests? :)

glye:
Added test.

glye:
The test failure is irrelevant: "Methods with the same name as their class will not be constructors in a future version of PHP; eZSiteAccess has a deprecated constructor"
This is in legacy itself, not the LegacyBridge. And it affects PHP 7+ only.

joaoinacio:
My 2 cents: Looks OK, but a bit complex to me as i'm no expert.

Not really certain how the fragment is/should be handled, but I would imagine something simpler such as the following (plus utf handling):
```php
    if ($pathinfo != $semanticPathinfo) {
            if ( $p = strpos($pathinfo, $semanticPathinfo) ) {
                $uriPart = substr($pathinfo, 0, $p );
                $uriPart = explode('/', substr($uriPart, 1));
            }
    }
```

glye:
@joaoinacio Such a simpler approach may be a good idea, but I think your example can fail because it uses strpos to search from the beginning of the string - we need to search from the end to avoid matching too much, in certain cases.

glye:
Update: Simpler version using mb_strrpos.

andrerom:
> The failing test is irrelevant, same as before.

It's relevant, see way to avoid it here: https://github.com/ezsystems/LegacyBridge/pull/53/commits/da8e9506b4559640d7168c76956efe39b2bb49ce

glye:
@andrerom Thanks! Updated test to make php7 pass, and rebased on master now that https://jira.ez.no/browse/EZP-25877 is merged.

glye:
Tests passing. Review ping @andrerom @joaoinacio @yannickroger

joaoinacio:
Looks OK AFAICT, can't really say about the fragment handling but there is a green light from tests so... :)

bdunogier:
Could you give an example (with actual URIs) of how your change will behave ? To be honest, it is hard to give a meaningful advice without going really deep into it.

glye:
@bdunogier Improved main description with URI example.

bundle/LegacyMapper/SiteAccess.php
```diff
@@ -81,11 +81,16 @@ public function onBuildKernelWebHandler(PreBuildKernelWebHandlerEvent $event)
...
-            $aSemanticPathinfo = explode('/', substr($semanticPathinfo, 1));
-            $uriPart = array_diff($aPathinfo, $aSemanticPathinfo, $this->getFragmentPathItems());
+        $fragmentPathInfo = implode('/', $this->getFragmentPathItems());
+        if ($fragmentPathInfo !== '_fragment') {
+            $semanticPathinfo = $fragmentPathInfo;
+        } else {
+            $semanticPathinfo = $request->attributes->get('semanticPathinfo', $pathinfo);
+        }
+        if ($pathinfo != $semanticPathinfo &&
+            $pos = mb_strrpos($pathinfo, $semanticPathinfo)) {
```
glye: Done.

glye:
Update: Dealing with string not found, and visual cleanup.

glye:
@bdunogier You cool with this now?

bundle/LegacyMapper/SiteAccess.php
```diff
@@ -81,11 +81,16 @@ public function onBuildKernelWebHandler(PreBuildKernelWebHandlerEvent $event)
...
-            $uriPart = array_diff($aPathinfo, $aSemanticPathinfo, $this->getFragmentPathItems());
+        $fragmentPathInfo = implode('/', $this->getFragmentPathItems());
+        if ($fragmentPathInfo !== '_fragment') {
+            $semanticPathinfo = $fragmentPathInfo;
+        } else {
+            $semanticPathinfo = $request->attributes->get('semanticPathinfo', $pathinfo);
+        }
+        $pos = mb_strrpos($pathinfo, $semanticPathinfo);
+        if ($pathinfo != $semanticPathinfo && $pos !== false) {
+            $uriPart = mb_substr($pathinfo, 0, $pos);
```
bdunogier: Could these two `mb_substr` calls be changed to one ? We could use `1` and `$pos` couldn't we ?

bdunogier:
+1 aside from the `mb_substr` nitpick.

bundle/LegacyMapper/SiteAccess.php
glye: Yep, seems they can.

glye:
I just realised I should have opened this against ezpublish-kernel-ee 5.3/5.4 and forward ported to LegacyBridge. Since it's approved I'll just push it there, and port from there.

glye:
Closing. This was merged in https://github.com/ezsystems/LegacyBridge/commit/d30373f67c3ee339aa64946e9533b275351d0746 which was forward-ported from the right source.

PR #63, opened 2016-09-23
Fix fatal errors with Twig 1.25.0 and Twig 2.0

emodric:
As of Twig 1.25.0, there's a fatal error:

> Fatal error: Class eZ\Publish\Core\MVC\Legacy\Templating\Twig\Template contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (Twig_Template::getDebugInfo)

Caused by https://github.com/twigphp/Twig/pull/2132

@andrerom Can you make a patch release ASAP, please?

EDIT: Also added a second method to make sure there's no fatal error in Twig 2.0

### DIFF
--- a/mvc/Templating/Twig/Template.php
+++ b/mvc/Templating/Twig/Template.php
@@ -66,6 +66,14 @@ class Template extends Twig_Template
     /**
      * {@inheritdoc}
      */
+    public function getDebugInfo()
+    {
+        return array();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
     protected function doDisplay(array $context, array $blocks = array())
     {
     }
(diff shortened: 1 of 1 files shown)


emodric:
Ping @andrerom

lolautruche:
+1

flovntp:
+1

lolautruche:
Ping @bdunogier

emodric:
@bdunogier Thanks for the merge! Can you also tag, please?

bdunogier:
Is master ripe enough for a v1.0.4 ? (alternative is a v1.0.3.1).

emodric:
Is there a difference? I mean, most probably people use legacy bridge with `^1.0` or `~1.0`, so it won't make any difference.

lolautruche:
I'd say `v1.0.4`. There are a few interesting fixes since `v1.0.3`

bdunogier:
The difference is that if tag a 1.0.3.1, it will only contain this change :)

If I tag a v1.0.4, it will contain the current master.

PR #64, opened 2016-09-29
Use FQCN in Twig_Extension::getName

emodric:
`Twig_Extension::getName` is deprecated in Twig 1.26 (https://github.com/twigphp/Twig/pull/2148) and removed in Twig 2.0.

Using FQCN in `getName` makes sure the extension keeps functioning in Twig 1.x as well as 2.0, where FQCN is used to reference the extensions.

### DIFF
--- a/mvc/Templating/Twig/Extension/LegacyExtension.php
+++ b/mvc/Templating/Twig/Extension/LegacyExtension.php
@@ -84,7 +84,7 @@ class LegacyExtension extends Twig_Extension implements Twig_Extension_InitRunti
      */
     public function getName()
     {
-        return 'ezpublish.legacy';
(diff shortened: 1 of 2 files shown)


andrerom:
Will this work on older twig versions or should we bump twig requirement?

emodric:
There's no reason why it shouldn't work. On older Twig versions, extension name is just a string.

The only potential breaking change would be if someone directly used `Twig_Environment::getExtension` method, with the old name (`ezpublish.legacy`). But in my option, it is not likely that someone uses `Twig_Environment` like that.

lolautruche:
+1

andrerom:
+1

PR #66, opened 2016-10-06
Add vhost config to install instructions

emodric:
Together with https://github.com/ezsystems/LegacyBridge/pull/65, this is something that people tend to miss when installing legacy bridge


andrerom:
Needs update now ;)

emodric:
Updated!

PR #68, opened 2016-10-06
Fix EZP-26389: EZP-26181 truncates uri_part

wizhippo:
https://jira.ez.no/browse/EZP-26389


glye:
Ping @joaoinacio @bdunogier @andrerom

PR #70, opened 2016-10-10
Fix EZP-26393: Missing options for ezpublish_legacy.kernel_handler.cli

lolautruche:
> https://jira.ez.no/browse/EZP-26393

This patch enforces `use-modules` to `true`, so that operations like `content/publish` can be used from legacy scripts:

```php
eZOperationHandler::execute('content', 'publish', [
    'object_id' => $objectId, 
    'version' => $version,
]);
```

### DIFF
--- a/bundle/Resources/config/services.yml
+++ b/bundle/Resources/config/services.yml
@@ -26,7 +26,8 @@ parameters:
     ezpublish_legacy.website_toolbar.controller.class: eZ\Bundle\EzPublishLegacyBundle\Controller\WebsiteToolbarController
 
     ezpublish_legacy.kernel_handler.cli.class: eZ\Publish\Core\MVC\Legacy\Kernel\CLIHandler
-    ezpublish_legacy.kernel_handler.cli.options: {}
+    ezpublish_legacy.kernel_handler.cli.options:
+        use-modules: true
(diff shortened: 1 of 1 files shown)


lolautruche:
Ping @andrerom @bdunogier @yannickroger @glye

emodric:
+1

Why not include `use-extensions` too?

lolautruche:
> Why not include use-extensions too?

Because it's `true` by default 😉

emodric:
> Because it's true by default :wink:

Well, damn! :)

glye:
Not knowing this stuff, is this just something that has been forgotten, or are there any downsides/tradeoffs?
The reason for not changing the use-modules default value instead is... BC?

lolautruche:
Yes, when implementing the CLIHandler, I actually forgot an important usecase: operations, which are bound on modules.

I remember wondering why would we need modules in a script... Stupid me.

lolautruche:
There are no tradeoffs AFAIK

bdunogier:
+1

lolautruche:
If everyone approves, maybe this can be merged? :-)

lolautruche:
Ping ? :)

PR #74, opened 2016-11-21
Modify permissions folder check.

raupie:
CheckList was using legacy ezpublish folder name in relative path. Updated CheckList with new app folder name. A warning was being thrown during the ezsetup wizard when trying to install the legacybundle overtop of ezplatform.


emodric:
+1 from me. AFAIK, there is no version of eZ Platform that uses `ezpublish/cache` et al.

Ping @bdunogier @crevillo

crevillo:
+1

PR #75, opened 2016-11-29
Fix the SSO Handler connector

Plopix:
Not sure since when we have to do that, I think it is probably Symfony 2.8


andrerom:
review ping @emodric @lolautruche

lolautruche:
+1

emodric:
+1

PR #77, opened 2017-01-12
Allow sensio/distribution-bundle ^5.0

emodric:
`composer install` works fine with 5.0 too.


emodric:
Huh! `ConfigurationDumper` still uses `\Sensio\Bundle\DistributionBundle\Configurator\Configurator` which was removed in 5.0...

It looks like it writes a random secret to `app/config/parameters.yml` after legacy setup wizard is ran.

I think it's safe to remove it: https://github.com/ezsystems/LegacyBridge/blob/master/bundle/SetupWizard/ConfigurationDumper.php#L101-L108

@bdunogier WDYT?

emodric:
@wizhippo Your input is appreciated too :)

wizhippo:
@emodric I believe that should be able to be removed too with no issue as when you use composer to install you are asked to set the secret.  I see no need to reset the secret again from the legacy setup wizard.

bdunogier:
I agree with you guys.

emodric:
PR updated to remove the configurator.

andrerom:
Ping @lolautruche

andrerom:
From my side +1 but wanted JV's pov

lolautruche:
Shouldn't we regenerate the secret?

Other than that, +1

emodric:
Why though?

Initial composer install asks the user to set the secret. Why should legacy bridge install regenerate user set secret?

bdunogier:
> Initial composer install asks the user to set the secret. Why should legacy bridge install regenerate user set secret?

I think I agree with that.

lolautruche:
Alright then :-)

andrerom:
v1.1.0 version for this maybe then? Other feature changes we should aim for while at it?

lolautruche:
I'd say `1.1.0` because it can be a blocker sometimes.

emodric:
`1.1.0` sounds okay. I'd like to push another PR to fix deprecations in YML, after merging this, but before tagging.

lolautruche:
Can this be backported to 5.4 please?

DistributionBudle 3.x is way too old and is causing issues with composer such as:

```
  PHP Warning:  Module 'sysvshm' already loaded in Unknown on line 0
  PHP Warning:  Module 'tokenizer' already loaded in Unknown on line 0
  PHP Warning:  Module 'wddx' already loaded in Unknown on line 0
  PHP Warning:  Module 'xmlreader' already loaded in Unknown on line 0
  PHP Warning:  Module 'xmlwriter' already loaded in Unknown on line 0
  PHP Warning:  Module 'xsl' already loaded in Unknown on line 0
  PHP Warning:  Module 'zip' already loaded in Unknown on line 0
```

Thanks :-)

emodric:
@lolautruche Latest Distribution bundle 4.x works fine :)

lolautruche:
OK I'll check this out, thanks @emodric

PR #79, opened 2017-02-02
Ignore legacy mapper exceptions when there is no database

emodric:
The exception thrown here breaks `composer install` when there is no database, for example, when doing an initial installation of a repo which includes the legacy bridge.

### DIFF
--- a/bundle/LegacyMapper/Configuration.php
+++ b/bundle/LegacyMapper/Configuration.php
@@ -21,6 +21,7 @@ use ezxFormToken;
 use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use RuntimeException;
+use Exception;
 
 /**
  * Maps configuration parameters to the legacy parameters.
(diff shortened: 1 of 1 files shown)


andrerom:
No candidates for more specifc catch?

emodric:
> No candidates for more specifc catch?

Other than gracefully handling it in the kernel itself, no, not really.

emodric:
Free to merge? :)

andrerom:
last review needed, ping @glye @bdunogier

emodric:
Yeah, the only time I ran into the issue was when database was not available and it only happened on first composer install of a project that has legacy bridge preinstalled.

What happens is a straight SQL execution exception, so it is not handled by eZ kernel in any way.

I can add logging though :)

andrerom:
So then this change could have catched that exception instead or? (that is what both @glye and me are saying we would prefer here)

glye:
There's an example in the comments here for catching `mysqli_sql_exception` for example: http://php.net/manual/en/class.mysqli-sql-exception.php
However then you'd need to take other drivers into account, too.

emodric:
I know what you mean @andrerom and @glye, I'm not a fan of use of global exceptions, but as I said, since this happens in a very specific situation, where there is no database present and nowhere else, there is not much left to do.

One more specific error to catch would be `RuntimeException` (since that's the exception that `ExceptionConversion` gateways throw), but it's basically the same as using `Exception`.

glye:
Right. `ExceptionConversion` does set the title `Database error`, which you *could* check for, and re-throw the exception if it isn't set, but that's possibly even uglier ;)

emodric:
> which you could check for, and re-throw the exception if it isn't set, but that's possibly even uglier ;)

It is :D :skull_and_crossbones:

andrerom:
true, exception conversion :|

ok, then merge it is then.

emodric:
@andrerom Thanks :)

Now that this is out of the way, can I tag `1.1.1` ?

andrerom:
Feel free :)

emodric:
:+1:

PR #80, opened 2017-02-09
Depend on interfaces in view providers instead of concrete classes

emodric:
Backwards compatible change that enables use of these providers with custom View objects.


emodric:
Ready for merge then? :)

emodric:
Thanks! :)

PR #82, opened 2017-02-25
Add support for clean legacy install

yannickmodahgouez:
This adds support for a CLI installation of legacy clean content using : 

`php app/console ezplatform:install legacy_clean`

### DIFF
--- a/bundle/Installer/CleanInstaller.php
+++ b/bundle/Installer/CleanInstaller.php
@@ -1,0 +1,27 @@
+<?php
+
+namespace eZ\Bundle\EzPublishLegacyBundle\Installer;
+
+class CleanInstaller extends DbBasedInstaller implements Installer
+{
+    public function createConfiguration()
+    {
+    }
+
+    public function importSchema()
+    {
+        $this->runQueriesFromFile('ezpublish_legacy/kernel/sql/mysql/kernel_schema.sql');
+        $this->runQueriesFromFile('ezpublish_legacy/kernel/sql/mysql/cluster_dfs_schema.sql');
+    }
+
+    public function importData()
+    {
+        $this->runQueriesFromFile(
+            'ezpublish_legacy/kernel/sql/common/cleandata.sql'
+        );
+    }
+
+    public function importBinaries()
+    {
+    }
+}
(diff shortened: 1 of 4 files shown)


yannickmodahgouez:
This doesn't work with the current sql files in ezpublish legacy kernel since they are not compatible with ezplatform schema. There should be another set of files somewhere in a data folder. But should it be in the legacy kernel or in the legacy bridge ? 🤔

andrerom:
Hi @yannickmodahgouez :)

What's the schema error?

yannickmodahgouez:
Hey @andrerom, there is no schema error per se, ezplaform doesn't behave the way it should when I use the schema / data stored in ezpublish_legacy kernel. I'll try to dig up to give you more info.

emodric:
I never had any problems with using just `ezplatform:install clean` :)

Also, this does complicate maintenance, since changes to original schema will need to be applied here too.

yannickmodahgouez:
Well my main issue with ezplatform:install clean is that it deploys content that is not fully compatible with the legacy ( ezxmlrichtext ) and it might be confusing for users

emodric:
Hah, okay, I didn't think abot that one :) I usually remove the fields from Platform UI and don't give it any more thought.

yannickmodahgouez:
@emodric yes I did the same but I figured out it would be better to have a more straightforward install process.

@andrerom I confirm that there is no issue with the schema, ezplatform was acting up because the ezxmltext converter was missing, I required ezplatform-xmltext-fieldtype and everything is fine now. Well except the fact that some of the content types added are really old ( ini settings, Template look ... ). I think they were removed during the 5.x era aren't they ?
Also ideally, there should be some DROP TABLE IF EXISTS statements in the schemas. But that's just me being picky here.

andrerom:
@yannickmodahgouez So then I would suggest you consider extending the clean install and reuse the schema function, and only do custom data.

emodric:
+1 on @andrerom 's comment :)

andrerom:
> Btw, for making the richtext fields usable on the raw level there is "netgen/richtext-datatype-bundle"

Yes, proposed doc to reflect that in #119 also now :)

robinmuilwijk:
If I'm correct, this pull request would make legacy install simpler. If this is the case, @bdunogier, @andrerom and @yannickmodahgouez , how can we get this pull request moving? Thanks.

andrerom:
@robinmuilwijk the pr needs _lots_ of cleanup, so either @yannickmodahgouez or someone in community would need to work on it.

PR #83, opened 2017-02-25
Add support for ezxmltext field type

yannickmodahgouez:


andrerom:
I guess this should be separate commit on your db install PR, as that is the one adding the requirement.

yannickmodahgouez:
@andrerom damn I just moved it to another branch on purpose (:

andrerom:
both works of course, so if it's is ok like this with @emodric then we can leave it like this.

emodric:
Ok with me!

andrerom:
Closing as the change was moved to the other PR :)

PR #54, opened 2016-04-26
EZP-25720: Update type hint to use the proper event class

wizhippo:
See https://github.com/sensiolabs/SensioDistributionBundle/commit/964a56e855acac38d4a81920b3a86543f7e8492f

PR #86, opened 2017-03-23
[Composer] Make sure we require supported versions of sensio/distribu…

andrerom:
…tion-bundle

Ref: https://github.com/ezsystems/LegacyBridge/pull/54#issuecomment-288666609


emodric:
+1

PR #88, opened 2017-04-13
EZP-27249: Fix routerpass to use defined class

wizhippo:
Follow kernel and no longer use router.class parameter

### DIFF
--- a/bundle/DependencyInjection/Compiler/RoutingPass.php
+++ b/bundle/DependencyInjection/Compiler/RoutingPass.php
@@ -8,6 +8,7 @@
  */
 namespace eZ\Bundle\EzPublishLegacyBundle\DependencyInjection\Compiler;
 
+use eZ\Bundle\EzPublishLegacyBundle\Routing\DefaultRouter as DefaultRouter;
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 
@@ -21,6 +22,10 @@ class RoutingPass implements CompilerPassInterface
         if (!$container->hasDefinition('router.default')) {
             return;
         }
+
+        $container
+            ->findDefinition('router.default')
+            ->setClass(DefaultRouter::class);
 
         $defaultRouterDef = $container->getDefinition('router.default');
         $defaultRouterDef->addMethodCall(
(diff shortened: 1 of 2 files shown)


emodric:
@wizhippo You tested and it works?

If so, +1.

bundle/DependencyInjection/Compiler/RoutingPass.php
```diff
@@ -22,6 +23,10 @@ public function process(ContainerBuilder $container)
             return;
         }
 
+        $container
+            ->findDefinition('router.default')
+            ->setClass(DefaultRouter::class);
```
emodric: This should use a FQCN as string, since legacy bridge still supports PHP 5.4.

emodric:
@wizhippo Now I see there's already `$defaultRouterDef` variable just below your code. You can then add the `setClass` call to that variable instead of using `findDefinition` :)

wizhippo:
@emodric Working here so far.

bundle/Resources/config/services.yml
```diff
@@ -6,7 +6,6 @@ parameters:
     ezpublish.default_router.legacy_aware_routes: ['_ezpublishLegacyTreeMenu', 'ezpublish_rest_', '_ezpublishPreviewContent', '_wdt', '_profiler', '_assetic']
 
     # Core overrides
-    router.class: eZ\Bundle\EzPublishLegacyBundle\Routing\DefaultRouter
     ezpublish.security.login_listener.class: eZ\Bundle\EzPublishLegacyBundle\Security\SecurityListener
     security.authentication.listener.rememberme.class: eZ\Bundle\EzPublishLegacyBundle\Security\RememberMeListener
```
andrerom: if this comes from symfony, then this is broken on 3.x as well.

emodric: Yep. We can fix it in separate PR?

wizhippo: Are you suggesting another PR just for the parameter removal?

Not quite sure I follow.  As the compiler pass now sets the class, this parameter is not used anyway.

Let me know what you want done.

emodric: @wizhippo I think @andrerom is referring to `security.authentication.listener.rememberme.class` parameter which comes from Symfony, so Legacy Bridge will not work on Symfony 3.x because this parameter was removed.

andrerom: yes, it can be done in separate PR, just noticed it while reviewing here ;)

emodric:
Thanks @wizhippo 👍

PR #89, opened 2017-04-13
PHP CS Fixer v2

emodric:
Added PHP CS Fixer v2 config and ran the fixer.

### DIFF
--- a/bundle/LegacyResponse/LegacyResponseManager.php
+++ b/bundle/LegacyResponse/LegacyResponseManager.php
@@ -6,6 +6,7 @@
  * @license For full copyright and license information view LICENSE file distributed with this source code.
  * @version //autogentag//
  */
+
 namespace eZ\Bundle\EzPublishLegacyBundle\LegacyResponse;
 
 use eZ\Bundle\EzPublishLegacyBundle\LegacyResponse;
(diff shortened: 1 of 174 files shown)


bundle/LegacyResponse/LegacyResponseManager.php
```diff
@@ -177,7 +178,7 @@ public function mapHeaders(array $headers, Response $response)
                 case 'expires':
                     $response->setExpires(new DateTime($headerValue));
                     break;
-                default;
+                default:
```
andrerom: bug fix?

emodric: Came from the fixer automatically (probably by allowing the risky fixers), but since tests passed decided to push it.

andrerom: ok

emodric:
```php
<?php

$var = 2;

switch ($var) {
    case 1:
        var_dump(1);
        break;
    default;
        var_dump('default');
        break;
}
```
```
eddie@abyss [~] $ php test.php
string(7) "default"
```

Issue #87, opened 2017-04-08
ezpublish_legacy handle Not foud error (404)


kemoc:
in my configuration
ezplatform.yml
```
ez_publish_legacy:
    #enabled: false
    system:
        ztest:
            legacy_mode: false
```
routing.yml
```
#...
_ezplatformRepositoryFormsRoutes:
    resource: "@EzSystemsRepositoryFormsBundle/Resources/config/routing.yml"

_contentOnTheFly:
    resource: "@ContentOnTheFlyBundle/Resources/config/routing.yml"
    prefix:   '%ezpublish_rest.path_prefix%'

_ezpublishLegacyRoutes:
    resource: "@EzPublishLegacyBundle/Resources/config/routing.yml"

#<end of file>
```
My eZPlatform is v1.8.1 but this problem is also on v1.3.x

emodric:
@kemoc Can you please describe the issue you're having in more detail? It's not really clear what the problem is.

kemoc:
Hi @emodric I did bridge Readme.md and i disabled legacy for some siteaccess and i am surprised ezpublish 4 shows error 404 but should do it Symfony stack. For enabled legacy for other siteaccess in YML config legacy shows 404 too. I got whole layout from legacy (pagelayout + content of error module/view)

kemoc:
Hi @emodric I did bridge Readme.md and i disabled legacy for some siteaccess and i am surprised ezpublish 4 shows error 404 but should do it Symfony stack. For enabled legacy for other siteaccess in YML config legacy shows 404 too. I got whole layout from legacy (pagelayout + content of error module/view)

emodric:
This is due to the fact that legacy bridge does not convert legacy 404 errors into Symfony exceptions, only 401 and 403:

https://github.com/ezsystems/LegacyBridge/blob/master/bundle/LegacyResponse/LegacyResponseManager.php#L87-L100

I'm not sure why is that, maybe @lolautruche can shed some light?

lolautruche:
There is no conversion because there is no need, as `$response->setStatusCode()` is called with the error code provided by legacy kernel.

emodric:
@lolautruche Well, it means that you can't use Symfony 404 error templates since Symfony's `NotFoundHttpException` is never thrown and eZ Platform falls back to legacy error templates which is not ideal.

[...]

PR #91, opened 2017-04-24
Add Option to use NotFoundHttpException for 404 HTTP errors

blankse:
Issue: https://github.com/ezsystems/LegacyBridge/issues/87

- [X] Unit test
- [x] Manual test

### DIFF
--- a/bundle/LegacyResponse/LegacyResponseManager.php
+++ b/bundle/LegacyResponse/LegacyResponseManager.php
@@ -14,6 +14,7 @@ use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 use Symfony\Component\Templating\EngineInterface;
 use DateTime;
@@ -47,6 +48,13 @@ class LegacyResponseManager
     private $legacyMode;
 
     /**
+     * Flag indicating if we're convert 404 errors into a NotFoundHttpException.
+     *
+     * @var bool
+     */
+    private $notFoundHttpConversion;
+
+    /**
      * @var RequestStack
      */
     private $requestStack;
@@ -55,6 +63,7 @@ class LegacyResponseManager
     {
         $this->templateEngine = $templateEngine;
         $this->legacyLayout = $configResolver->getParameter('module_default_layout', 'ezpublish_legacy');
+        $this->notFoundHttpConversion = $configResolver->getParameter('not_found_http_conversion', 'ezpublish_legacy');
         $this->legacyMode = $configResolver->getParameter('legacy_mode');
         $this->requestStack = $requestStack;
     }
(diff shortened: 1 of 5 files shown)


emodric:
Looks good to me :+1:

blankse:
@kemoc @lolautruche Can you review this?

kemoc:
@blankse seems legit

lolautruche:
This should be documented somehow.
I'm not a fan of having a setting for that but I don't see another way to keep BC without it.

Any idea @bdunogier @andrerom ?

andrerom:
@lolautruche well, I guess the new behavior is what is should be, it's after all non legacy mode. What about introducing that as the default and only behavior as of 1.3.0?

lolautruche:
> What about introducing that as the default and only behavior as of 1.3.0

I don't agree, since this would be a BC break. This should go to 2.0.
(And the fix should be backported to 5.4 IMHO)

emodric:
I'm with @lolautruche on this one.

emodric:
What we can do is add an `E_USER_DEPRECATED` in case legacy kernel handles the exception.

bundle/LegacyResponse/LegacyResponseManager.php
```diff
@@ -87,11 +97,27 @@ public function generateResponseFromModuleResult(ezpKernelResult $result)
...
+                    $errorMessage = isset($moduleResult['errorMessage']) ? $moduleResult['errorMessage'] : 'Access denied';
+                    throw new AccessDeniedException($errorMessage);
+                }
+                // If having an "Not found" error code in non-legacy mode and conversation is true,
+                // we send an NotFoundHttpException to be able to trigger error page in Symfony stack.
+                if ($moduleResult['errorCode'] == 404) {
+                    if ($this->notFoundHttpConversion) {
+                        $errorMessage = isset($moduleResult['errorMessage']) ? $moduleResult['errorMessage'] : 'Not found';
+                        throw new NotFoundHttpException($errorMessage);
+                    } else {
```
c-revillo: is the `else` needed? i mean, you are throwing the exception so maybe else can be avoided...

blankse: You are right. I removed the else ;)

c-revillo: :+1:

blankse:
@emodric I added an E_USER_DEPRECATED if legacy handles the 404 error.

bundle/LegacyResponse/LegacyResponseManager.php
```diff
@@ -106,9 +106,16 @@ public function generateResponseFromModuleResult(ezpKernelResult $result)
...
-                    $errorMessage = isset($moduleResult['errorMessage']) ? $moduleResult['errorMessage'] : 'Not found';
-                    throw new NotFoundHttpException($errorMessage);
+                if ($moduleResult['errorCode'] == 404) {
+                    if ($this->notFoundHttpConversion) {
+                        $errorMessage = isset($moduleResult['errorMessage']) ? $moduleResult['errorMessage'] : 'Not found';
+                        throw new NotFoundHttpException($errorMessage);
+                    }
+                    @trigger_error(
+                        "Legacy 404 error handling is deprecated, and will be removed in legacy-bridge 2.0.\n" .
+                        'Use the not_found_http_conversion setting to use the new behavior yet.',
```
emodric: `to use the new behavior and disable this notice`

emodric:
@blankse +1 With an inline nitpick fixed

bundle/LegacyResponse/LegacyResponseManager.php
blankse: I changed it

PR #92, opened 2017-05-04
Add overriden NotFoundHttpException to have access to the original response

emodric:
Now that #91 is done, it would greatly help to know what happened in legacy when a 404 error is received.

We have some siteaccesses that run in legacy mode set to false, but still display legacy 404 errors, so this would help facilitate that.

It is worthwhile to note that this behaviour (this custom 404 exception) would still be part of 2.0 release.

### DIFF
--- a/bundle/LegacyResponse/LegacyResponseManager.php
+++ b/bundle/LegacyResponse/LegacyResponseManager.php
@@ -14,7 +14,7 @@ use eZ\Publish\Core\MVC\ConfigResolverInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use eZ\Bundle\EzPublishLegacyBundle\Exception\NotFoundHttpException;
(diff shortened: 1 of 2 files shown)


emodric:
Ping @andrerom @blankse

bundle/LegacyResponse/LegacyResponseManager.php
```diff
@@ -104,13 +104,18 @@ public function generateResponseFromModuleResult(ezpKernelResult $result)
...
                 }
-                // If having an "Not found" error code in non-legacy mode and conversation is true,
+
+                // If having an "Not found" error code in non-legacy mode and conversion is true,
                 // we send an NotFoundHttpException to be able to trigger error page in Symfony stack.
                 if ($moduleResult['errorCode'] == 404) {
                     if ($this->notFoundHttpConversion) {
                         $errorMessage = isset($moduleResult['errorMessage']) ? $moduleResult['errorMessage'] : 'Not found';
-                        throw new NotFoundHttpException($errorMessage);
+                        $exception = new NotFoundHttpException($errorMessage);
```
andrerom: can't you use custom ctor given it is custom exception?

emodric: Sure, done!

PR #93, opened 2017-05-29
hasInstance method is missing in the Kernel.php

quentinthai:
Hello,

We're trying to migrate our former sites developed with eZ Publish 5.x in Legacy mode to eZ Platform 6.9.x with LegacyBridge 1.2.x.

I've noticed that the following folder has been moved from eZ Publish 5.x: 
**vendor/ezsystems/ezpublish-kernel/eZ/Publish/Core/MVC/Legacy** 
to 
**vendor/ezsystems/legacy-bridge/mvc**

And the method **hasInstance()** has been removed from the class **Kernel.php**.

Could you give us the reason? Is it because it was used for Singleton Pattern which is considered now as an anti-pattern?

And is it possible to reintegrate it again?

The issue is that method is called in the following class: 
**vendor/ezsystems/legacy-bridge/bundle/Collector/TemplateDebugInfo.php**

And other various classes as well.

Thank you in advance for your feedback.

Regards,

Quentin

Here is the method hasInstance extracted from eZ Publish 5.x :

>     /**
>      * Checks if LegacyKernel has already been instantiated.
>      *
>      * @return bool
>      */
>     public static function hasInstance()
>     {
>         return static::$instance !== null;
>     }

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /vendor/
 composer.lock
 ezpublish_legacy
+var
 bin
 config.php
 .php_cs.cache
--- a/.php_cs
+++ b/.php_cs
@@ -1,24 +1,21 @@
 <?php
-
-return Symfony\CS\Config\Config::create()
-    ->setUsingLinter(false)
-    ->setUsingCache(true)
-    ->level(Symfony\CS\FixerInterface::SYMFONY_LEVEL)
-    ->fixers([
-        'concat_with_spaces',
-        '-concat_without_spaces',
-        '-empty_return',
-        '-phpdoc_params',
-        '-phpdoc_separation',
-        '-phpdoc_to_comment',
-        '-spaces_cast',
-        '-blankline_after_open_tag',
-        '-single_blank_line_before_namespace',
-        // psr0 has weird issues with our PSR-4 layout, so deactivating it.
-        '-psr0',
+return PhpCsFixer\Config::create()
+    ->setRules([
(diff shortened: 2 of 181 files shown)


lolautruche:
Hi @quentinthai 
I think you need to rebase your branch :smiley:

quentinthai:
Hi @lolautruche 
No problem. I've just select the latest stable branch and put my PR in it... ^^

lolautruche:
I don't see your commit in the commits list

lolautruche:
@quentinthai This method has been removed in eb717c1dab2ccede8cd580c48eee27f86479f9db but should still work since `Kernel` extends `ezpKernel` (from ezpublish-legacy) which does have this class.

Can you please show the error + stack trace you get?

quentinthai:
@lolautruche, sure, here is my stack trace :

`Whoops, looks like something went wrong.

1/1
UndefinedMethodException in TemplateDebugInfo.php line 29:
Attempted to call an undefined method named "hasInstance" of class "eZ\Publish\Core\MVC\Legacy\Kernel".
in /home/users/webthai/ezplatform-poc-amundi/vendor/ezsystems/legacy-bridge/bundle/Collector/TemplateDebugInfo.php line 29`

quentinthai:
@lolautruche : And i don't have commit any code

quentinthai:
Moved to Issue: https://github.com/ezsystems/LegacyBridge/issues/94

PR #95, opened 2017-06-20
EZP 26013 cache update on trash restoration

tomaszmadeyski:
This is a (possible) fix for https://jira.ez.no/browse/EZP-26013 
This is a fix for https://jira.ez.no/browse/EZP-25240

## Description
For a full story check issue description and my [comment](https://jira.ez.no/browse/EZP-26013?focusedCommentId=211954&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-211954) specificaly.

What we're facing here is that when trying to move item out of trash a [content](https://github.com/ezsystems/LegacyBridge/blob/master/bundle/Cache/PersistenceCachePurger.php#L110) method is called with a set of `locationIds`. One of the ids is locationId (nodeId) of item after restoration from trash, so loading it both from [locationHandler](https://github.com/ezsystems/LegacyBridge/blob/master/bundle/Cache/PersistenceCachePurger.php#L129) fails. 

@pspanja in his [comment](https://jira.ez.no/browse/EZP-26013?focusedCommentId=201574&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-201574) suggested that clearing cache for content in general should help. I tested it and it indeed helps - hence this PR.

I'm not sure if this is only and best solution so I'm hoping for some opinions here

### DIFF
--- a/bundle/Cache/PersistenceCachePurger.php
+++ b/bundle/Cache/PersistenceCachePurger.php
@@ -11,6 +11,7 @@ namespace eZ\Bundle\EzPublishLegacyBundle\Cache;
 
 use Symfony\Component\HttpKernel\CacheClearer\CacheClearerInterface;
 use eZ\Publish\SPI\Persistence\Content\Location\Handler as LocationHandlerInterface;
+use eZ\Publish\SPI\Persistence\Content\Location;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentType;
 use eZ\Publish\Core\Base\Exceptions\NotFoundException;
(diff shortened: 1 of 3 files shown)


tomaszmadeyski:
Note: failing tests are not related with this PR

emodric:
Hm... This means that if we provide a "fake" location ID (the one which never existed), we get all content cache cleared? This is kinda misleading and not at all expected.

tomaszmadeyski:
Closing this in favour of https://github.com/ezsystems/ezpublish-legacy/pull/1306

PR #96, opened 2017-06-21
Added community badges

MarioBlazek:


emodric:
@andrerom Beat me to it by a fraction of second! :D

PR #98, opened 2017-07-13
Simplify install and add EE support with eZ Platform 1.11 + legacy 2017.08

andrerom:
Proposal for a ~~2.0.0~~ `1.3.0` release of legacy bridge, with aim to simplify installation. Implied in this simplification is also the need to provide a new legacy kernel release for use by community as well as potentially also eZ Enterprise users*. 

There might also be other things that can be simplified, suggestions (PR's) welcome :)

 \* _Afaik 5.4 legacy can't have features found in the 2017.08 milestone below needed for eZ Platform compatibility, and we don't intend to do a 5.5 release._ 

Bumps:
- legacy installer: in order to skip install step about defining `ezpublish-legacy-dir`
- legacy kernel: to be able to skip install steps about setting `EZP_APP_FOLDER_NAME`
- eZ Platform kernel _(ezpublish-kernel)_:  for future parity with 2017.08 legacy kernel
  - _Note: some of these features depends on community contributions, and might not make it._

TODO:
- [ ] Finalize and release legacy kernel [2017.08](https://github.com/ezsystems/ezpublish-legacy/milestone/1) and corresponding eZ Platform v1.11 release

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,8 +1,9 @@
 # Installing the eZ Platform legacy bridge
 
 Unlike eZ Publish 5.x, eZ Platform does not include the Legacy stack by default.
-Even though it is not officially supported, eZ Publish Legacy can easily be installed
-on top of Platform using Composer.
+
+Even though it is not recommended for use on new installs, eZ Publish Legacy can easily be installed
+on top of Platform using Composer to provide a more up-to-date platform to migrate your code to eZ Platform with.
 
 ### Missing legacy extensions
 
@@ -68,9 +69,7 @@ ezpublish_setup:
 
 ### Install `ezsystems/legacy-bridge`
 
-**Make sure you have set the ezpublish legacy folder in composer.json, as instructed above**
-
-`ezsystems/legacy-bridge` contains the libraries previous included in `ezsystems/ezpublish-kernel`.
+`ezsystems/legacy-bridge` contains the libraries previous included in `ezsystems/ezpublish-kernel` in version 5.x.
 
 It must be installed using Composer. Take care to use `^1.3.0` as version constraint, since previous versions lack some important fixes for newer versions of eZ Platform:
 
@@ -118,4 +117,4 @@ rewrite "^/var/storage/packages/(.*)" "/var/storage/packages/$1" break;
 
 Last step, if you are on *nix operation system, is to make sure to run 
 the appropriate command for setting correct folder permissions, you 
-can find the information you need in installation guide for eZ Publish 5.x.
+can find the information you need in [installation guide for eZ Publish 5.x](https://doc.ez.no/display/EZP/Installation).
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -66,40 +66,17 @@ ezpublish_setup:
     security: false
 ```
 
-### Configure ezpublish legacy's location in `composer.json`
-
-`ezsystems/ezpublish-legacy` needs to be installed in a particular location to work.
-
-Edit `composer.json`, and add `"ezpublish-legacy-dir": "ezpublish_legacy"` to the `extra` array:
-
-```
-    "extra": {
-        "ezpublish-legacy-dir": "ezpublish_legacy",
-```
-
 ### Install `ezsystems/legacy-bridge`
 
 **Make sure you have set the ezpublish legacy folder in composer.json, as instructed above**
 
 `ezsystems/legacy-bridge` contains the libraries previous included in `ezsystems/ezpublish-kernel`.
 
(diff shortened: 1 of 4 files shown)


emodric:
Is `2.0.0` really needed? I don't feel that this is a breaking change that warrants major version bump and we could do with `1.3.0`.

Other than that, :+1:

composer.json
```diff
@@ -11,8 +11,9 @@
     ],
     "minimum-stability": "stable",
     "require": {
-        "ezsystems/ezpublish-legacy": ">=2014.11",
-        "ezsystems/ezpublish-kernel": "~6.0@dev",
+        "ezsystems/ezpublish-legacy-installer": "^2.0.4",
+        "ezsystems/ezpublish-legacy": ">=2017.08",
```
emodric: Should this be `2017.07`?

andrerom: Nope, that was what I first called it, but if we aim for parity with the features in the milestone, then that needs work also on platform side, hence 1.11 => august => 08 :)

andrerom:
> and we could do with 1.3.0.

I'm fine with that.

emodric:
> if you or others have suggestions for what else we can do to further simplify the steps.

Not right now, no.

emodric:
> For instance could we add handling of legacy-installer to do some of the remaining steps? Among them enabling bundle and/or add post-update-cmd steps to composer.json

We can try, but then composer would need to be ran twice, to take into account the new bundle and new scripts, which IMO is worse for DX than simply adding those beforehand.

Plopix:
There are some settings to provide by default, to set up the Admin siteaccess etc.

bdunogier:
If we don't have other changes that don't require a major version, I'd rather go with a minor one. When we have identified and implemented the changes that require one, we can still do it.

But I have no strong feelings here and might be missing something.

Good idea in any case !

andrerom:
Changed to 1.3.0

> There are some settings to provide by default, to set up the Admin site access etc.

@Plopix What did you have in mind here? something you think we can add here by default somehow?

emodric:
> @Plopix What did you have in mind here? something you think we can add here by default somehow?

@andrerom Maybe something like this: https://github.com/emodric/ezplatform-legacy/tree/master/ezpublish_legacy/settings ?

INSTALL.md
```diff
@@ -66,41 +66,18 @@ ezpublish_setup:
...
-Edit `composer.json`, and add `"ezpublish-legacy-dir": "ezpublish_legacy"` to the `extra` array:
-
-```
-    "extra": {
-        "ezpublish-legacy-dir": "ezpublish_legacy",
-```
-
 ### Install `ezsystems/legacy-bridge`
 
 **Make sure you have set the ezpublish legacy folder in composer.json, as instructed above**
```
andrerom: needs to be removed, but done in #103

andrerom:
> @andrerom Maybe something like this: https://github.com/emodric/ezplatform-legacy/tree/master/ezpublish_legacy/settings ?

Anything in particular there you mean we should inject via the bridge, or implying we should trigger legacy config generation somehow?

emodric:
I don't think there's a need to generate anything. Nowdays, legacy bridge is used mostly for legacy admin, so just copying `settings/siteaccess/admin` and `settings/override` folders could be enough.

andrerom:
hmm, well opting for `admin` as admin SiteAccess might or might not conflict with v2 admin, and reason I mentioned generation was to make sure we take the right settings from platform.

emodric:
It can be done, sure. BTW, we're using Twig internally to do it when needed :)

Plopix:
Sorry long delay for this reply. Yes I was thinking about the settings for the Legacy Admin (basically everything we can't inject through the new stack)
Like here: https://github.com/Plopix/LegacyBridge/tree/feature-install-doc/settings_template it is probably too much, but it is usually what I add for legacy admin override and the admin site access.

emodric:
I would keep it optional in any case if done, not to collide with already existing configuration in existing sites.

andrerom:
passing now, I assume your approval still stands @emodric ? :)

INSTALL.md
```diff
@@ -66,41 +67,16 @@ ezpublish_setup:
...
-
-```
-    "extra": {
-        "ezpublish-legacy-dir": "ezpublish_legacy",
-```
-
 ### Install `ezsystems/legacy-bridge`
 
-**Make sure you have set the ezpublish legacy folder in composer.json, as instructed above**
+`ezsystems/legacy-bridge` contains the libraries previous included in `ezsystems/ezpublish-kernel` in version 5.x.
```
glye: previous*ly* included

PR #99, opened 2017-07-24
Fix EZEE-1656: Clarify that the legacy routes should be added at the end of the `app/config/routing` file

natanael89:
Related JIRA ticket: [EZEE-1656](https://jira.ez.no/browse/EZEE-1656)

Currently it is not specified where the legacy routes should be added to the `app/config/routing` file. If they are added before `_ezpublishDateBasedPublisher`, then the issue described in the JIRA ticket happens. To make sure it doesn't, the installation instructions should be clear that the legacy routes should be added after `_ezpublishDateBasedPublisher`.
For simplicity, I changed the installation instruction so it recommends adding the legacy routes at the end of the `app/config/routing` file.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -37,7 +37,7 @@ the `$bundles` array. Pay close attention to the `$this` argument. The LegacyBun
 of a spoiled brat, and has high expectations from its collaborators.
 
 ### Add legacy routes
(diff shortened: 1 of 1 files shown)


andrerom:
@natanael89 will do alternative PR to this making the api route pattern less greedy for 1.3.0 _(will need BC note in release if we go that way)_, might be the better fix for this.

PR #101, opened 2017-07-27
EZEE-1656: Further clarify route order

andrerom:
Make sure people have some inline reminder that legacy should be kept at the end of routes file.


emodric:
precedent > precendence
overrides > overriding

Otherwise :+1:

PR #102, opened 2017-07-27
[INSTALL] Add instructions for how to install with eZ Platform Enterprise

andrerom:

PR #103, opened 2017-07-31
Simplify install for 1.2 branch

andrerom:
- Bumps requirements to legacy-installer 2.0.4 & gets rid of instruction to configure legacy dir
- Updates kernel and legacy requirements so 1.3 will be needed for 6.11 / 2017.08 (#98)
- Add a cleaner note on how to install with 5.4 legacy then what was proposed in #102

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -78,6 +78,11 @@ It must be installed using Composer. Take care to use `^1.0.4` as version constr
 composer require --update-no-dev "ezsystems/legacy-bridge:^1.0.4"
 ```
 
+Alternatively, for eZ Platform EE users you should make sure to use the latest 5.4 version of legacy to receive full support:
+```
+composer require --update-with-dependencies "ezsystems/ezpublish-legacy:5.4.10" "ezsystems/legacy-bridge:^1.0.4"
+```
+
 ### Configuring Symfony app folder in legacy
 
 _Step not needed with version >= 5.4.7 (ee) and >= 2016.10 (community forks)_.
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -85,7 +85,7 @@ composer require --update-with-dependencies "ezsystems/ezpublish-legacy:5.4.10" 
 
 ### Configuring Symfony app folder in legacy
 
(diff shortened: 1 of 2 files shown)


andrerom:
@emodric It's not a must, not 100% certain myself. But idea was to make sure people needs the tentative LegacyBridge v1.3 release for 6.11 + 2017.08 and higher, to make sure composer gives proper errro if you are trying to run 6.11 and legacy 2014.11 for instance, or 6.7 and legacy 2017.08.

andrerom:
But on second thought composer might warn users anyway when legacy bridge 1.3 is out and their requirements on kernel or legacy does not match. I'll remove it.

emodric:
Maybe a better idea would be to add conflicts to legacy kernel?

andrerom:
@emodric Good point, that way it won't need to be a requirement also, so we can aim to handle that for 2017.08 then.

Updated this.

INSTALL.md
```diff
@@ -91,8 +78,15 @@ It must be installed using Composer. Take care to use `^1.0.4` as version constr
...
 ```
 
+Alternatively, for eZ Platform EE users you should make sure to use the latest 5.4 version of legacy to receive full support:
+```
+composer require --update-with-dependencies "ezsystems/ezpublish-legacy:5.4.10" "ezsystems/legacy-bridge:^1.0.4"
+```
+
 ### Configuring Symfony app folder in legacy
 
+_Step not needed with version >= 5.4.7 (ee) and >= 2016.10_.
```
emodric: 2016.10? You meant 2017.08 ?

andrerom: no, referring to fork releases here, from what I could see netgen fork for instance as of 2016.10 had the patch to autodetect app folder name, but can of course omit it if it just overcomplicates stuff here

emodric: If we're to tag legacy with 2017.08, there's no need to mention forks I think. Once official repo is tagged, I will add a note to the fork that it is deprecated and should not be used.

andrerom: ok

andrerom:
Closing in favor of just doing https://github.com/ezsystems/LegacyBridge/pull/98

PR #104, opened 2017-08-15
[Test] Move bundle test to right dir so it's executed with default config

andrerom:
Introduced in #133, this tests have been placed in wrong folder and is not tested by phpunit.


andrerom:
@glye any idea why the tests is failing? Asking since blame seems to indicate you added it.

emodric:
> @glye any idea why the tests is failing? Asking since blame seems to indicate you added it.

Whoops, didn't realize that it was failing :sweat: Vacation mode is still on I guess :D

glye:
Don't know yet. Odd it's passing in PHP 5.4. Obviously I was running these tests locally before merging, but back then it might have been only with PHP 5.4.

glye:
`$event->getParameters()->get('siteaccess[uri_part]', null, true)` returns null, where it should return the Admin uri part.
`$event->getParameters()` is a
    Symfony\Component\HttpFoundation\ParameterBag Object
        (
            [parameters:protected] => Array
                (
                    [siteaccess] => Array
                        (
                            [name] => Admin
                            [type] => 10
                            [uri_part] => Array
                                (
                                    [0] => Admin
                                )
                        )
                )
        )

Do you see anything obviously wrong here?

emodric:
@glye Third argument (`$deep = false`) was removed in Symfony 3.0, it seems, so it should be `$event->getParameters()->get('siteaccess')[uri_part]`.

glye:
Yep, I just saw that. Committed the fix.

emodric:
Thanks @glye :)

PR #106, opened 2017-09-04
EZP-27795: Notice thrown when declaring empty HTTP headers in Legacy

kmadejski:
JIRA issue: [https://jira.ez.no/browse/EZP-27795](https://jira.ez.no/browse/EZP-27795)

# Description
When a user creates own empty HTTP header (for e.g. *Pragma:*) using Legacy stack then notice `Undefined offset: 1` were thrown. This PR fixes this issue.


emodric:
Looks sane!

@kmadejski I presume your test was failing before the patch?

kmadejski:
@emodric yes, exactly as you said. If you revert these changes the test will fail.

emodric:
Thanks, merging now :)

PR #107, opened 2017-09-07
[Feature] Add command to setup legacy settings/designs symlinks

andrerom:
> This is done for doc effort in: https://github.com/ezsystems/developer-documentation/pull/28
> Which is from issue: https://jira.ez.no/browse/EZP-27142
> Part of Epic: https://jira.ez.no/browse/EZEE-1455

Without this feature, each and every user would have to find out the hard way that
Composer sometimes wipes out the `ezpublish_legacy` folder, and also that versioning
this folder is not really recommended so you somehow need to deal with settings and designs.

So this command setups and symlinks the following into legacy on compser install and update:
- src/legacy_files:
  - designs/*
  - settings/override
  - settings/siteaccess/*

The feature in this PR is primary aiming at upgrade/migration use cases, where
people are coming from 5.4 or 2014.x and similar and need to move over settings and designs.

Review wanted from partners that have their own way of doing this, as you will know
better what you need. /cc @emodric @joekepley @peterkeung ..

### Why no extension folder handling?
Extensions are handled by composer, and also by legacy bundles command by this bundle,
so if you need an extension in src folder you can make app bundle a legacy bundle.
Hence it felt easily confusing to give another option for this.

### Why no storage and storage/packages folder handling?
As site storage folder is not typically versioned in git, I guess this should be handled by something else then `src/`. Input on how people tend to deal with this is more then welcome!

### Why place them directly in `src/legacy_files`?
Opted for this mainly because placing it in bundle is more of a legacy bundles feature, and also makes sure we are forward compatible with Symfony 4/Flex's single bundle src layout.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,7 +15,7 @@ Edit `composer.json`, and add those lines to both `post-update-cmd` and `post-in
         "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installAssets",
         "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions",
         "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacySrcFiles"
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacyFiles"
     ],
     "post-install-cmd": [
         ...,
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -7,8 +7,8 @@ on top of Platform using Composer to provide a more up-to-date platform to migra
 
 ### Add the composer `legacy post-*-scripts`
 
-Edit `composer.json`, and add those lines to both `post-update-cmd` and `post-install-cmd` blocks at the end:
-
+Edit `composer.json`, and add those lines to both `post-update-cmd` and `post-install-cmd` blocks at the end, but before
+`eZ\Bundle\EzPublishCoreBundle\Composer\ScriptHandler::dumpAssets`:
 ```
 "scripts": {
     "legacy-scripts": [
@@ -27,6 +27,34 @@ Edit `composer.json`, and add those lines to both `post-update-cmd` and `post-in
     ],
 }
 ```
+
+Example: In the case of stock eZ Platform 1.11 and higher that would specifically be:
+```
+"scripts": {
+    "legacy-scripts": [
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installAssets",
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions",
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads",
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacyFiles"
+    ],
+    "symfony-scripts": [
+        "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
+        "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
+        "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::clearCache",
+        "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
+        "@legacy-scripts"
+        "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::dumpAssets",
+        "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile"
+    ],
+    "post-install-cmd": [
+        "@symfony-scripts"
+    ],
+    "post-update-cmd": [
+        "@symfony-scripts"
+    ],
+    (...)
+```
+
 
 ### Enable EzPublishLegacyBundle
 Edit `app/AppKernel.php`, and add `new eZ\Bundle\EzPublishLegacyBundle\EzPublishLegacyBundle( $this ),`
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -11,17 +11,19 @@ Edit `composer.json`, and add those lines to both `post-update-cmd` and `post-in
 
 ```
 "scripts": {
-    "post-install-cmd": [
-        ...,
+    "legacy-scripts": [
         "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installAssets",
         "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads"
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads",
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacySrcFiles"
+    ],
+    "post-install-cmd": [
+        ...,
+        "@legacy-scripts"
     ],
     "post-update-cmd": [
         ...,
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installAssets",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads"
+        "@legacy-scripts"
     ],
 }
 ```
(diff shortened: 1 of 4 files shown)


emodric:
I will go through the code in couple of days since I'm on a short break right now, but this looks okay to me.

At Netgen, we do something similar, with the difference that `legacy_files` is stored in a bundles' `Resources` folder, making it possible for every bundle to add it's own legacy stuff (i.e. legacy kernel patches). We also use different folders for different environments (e.g. `legacy_files` (used in every env), `legacy_files_dev`, `legacy_files_prod`) so we can have different configuration for different servers.

andrerom:
Cool, looking forward to it. Btw reason for opting for not putting legacy files in bundle was among other things because symfony 4 has single bundle layout afaik, and legacy bundles feature already allow you to do that.

brookinsconsulting:
+1

andrerom:
@emodric had a chance to have a look yet? :)

emodric:
@andrerom Nope, but will do it ASAP and let you know!

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -0,0 +1,150 @@
...
+        $srcArg = rtrim($input->getArgument('src'), '/');
+        $create = (bool)$input->getOption('create');
+        $force = (bool)$input->getOption('force');
+
+        /**
+         * @var \Symfony\Component\Filesystem\Filesystem
+         */
+        $filesystem = $this->getContainer()->get('filesystem');
+
+        if (!$filesystem->exists($srcArg)) {
```
emodric: What's the use creating the source folder structure?

INSTALL.md
```diff
@@ -11,17 +11,19 @@ Edit `composer.json`, and add those lines to both `post-update-cmd` and `post-in
...
 ```
 "scripts": {
-    "post-install-cmd": [
-        ...,
+    "legacy-scripts": [
         "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installAssets",
         "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads"
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads",
+        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacySrcFiles"
```
emodric: Maybe just `symlinkLegacyFiles` ?

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -0,0 +1,150 @@
...
+
+class LegacySrcSymlinkCommand extends ContainerAwareCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('ezpublish:legacy:symlink_src')
+            ->setDefinition(
+                array(
+                    new InputArgument('src', InputArgument::OPTIONAL, 'The src directory for legacy files', 'src/legacy_files'),
```
emodric: How about `src/legacy` as the default?

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -0,0 +1,150 @@
...
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+
+class LegacySrcSymlinkCommand extends ContainerAwareCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('ezpublish:legacy:symlink_src')
```
emodric: `ezpublish:legacy:symlink`?

(I'm not a fan of acronyms :P)

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -0,0 +1,150 @@
...
+            ->setDefinition(
+                array(
+                    new InputArgument('src', InputArgument::OPTIONAL, 'The src directory for legacy files', 'src/legacy_files'),
+                )
+            )
+            ->addOption('create', 'c', InputOption::VALUE_NONE, 'Create src folder structure if it does not exists')
```
emodric: `does not exist`

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -0,0 +1,150 @@
...
+     *
+     * @return array
+     */
+    protected function linkSrcFolders(Filesystem $filesystem, $srcArg, $force)
+    {
+        $symlinks = [];
+        $legacyRootDir = rtrim($this->getContainer()->getParameter('ezpublish_legacy.root_dir'), '/');
+
+        if ($force || !$filesystem->exists("$legacyRootDir/settings/override")) {
+            $filesystem->symlink(
```
emodric: Should probably check if `"$srcArg/settings/override"` exists before symlinking.

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -0,0 +1,150 @@
...
+                $filesystem->makePathRelative(
+                    realpath("$srcArg/settings/override"),
+                    realpath("$legacyRootDir/settings")
+                ),
+                "$legacyRootDir/settings/override"
+            );
+            $symlinks[] = "$legacyRootDir/settings/override";
+        }
+
+        $directories = ['design', 'settings/override'];
```
emodric: `settings/siteaccess` ?

emodric:
@andrerom Left a couple of remarks, but in general, this looks okay.

What I would like to definitely see is symlinking based on Symfony environment used when running the script (`src/legacy_files_dev`, `src/legacy_files_prod` and `src/legacy_files` (for all environments)).

bundle/Command/LegacySrcSymlinkCommand.php
andrerom: I was trying to not imply it should contain legacy itself, but if this is preferred or something else I'm open for it.

bundle/Command/LegacySrcSymlinkCommand.php
andrerom: Migration/Upgrade use case:
- install legacy bridge (will execute this step)
- manually move files over from old to new structure


So no manual steps needed for creating the folders, and it can be explicit referred to in doc where you move them as a default convention

bundle/Command/LegacySrcSymlinkCommand.php
andrerom: true, same for the rest

INSTALL.md
andrerom: If we rename folder name, yes

bundle/Command/LegacySrcSymlinkCommand.php
andrerom: yes, typo :)

peterkeung:
* "Composer sometimes wipes out the ezpublish_legacy folder" What are the conditions that cause this? (I'm pinging @ernestob.)

* We don't depend on Composer for deployments and don't use it on production environments. We use it only on dev to pull in code. So we do end up versioning ezpublish_legacy.

* Why the "design" folder? I guess people are still putting their templates here. We always put our templates in extensions.

emodric:
> Why the "design" folder? I guess people are still putting their templates here. We always put our templates in extensions.

Symlinking design folder is simpler than creating and activating the extension together with `design.ini.append.php`. For simpler usecases, symlinking designs to `ezpublish_legacy/design` is a welcome feature.

bundle/Command/LegacySrcSymlinkCommand.php
emodric: How hard is it to type `mkdir -p src/legacy_files/settings` ? :) Doesn't seem that much useful to me if it's documented right, but okay, if you think it can help, sure, but I still don't like that script is running with `-c` automatically *every time*.

andrerom:
> Why the "design" folder? I guess people are still putting their templates here. We always put our templates in extensions.

I assumed some people use it given it was never deprecated in favour of extensions, as for extensions they are as said already handled by other features of this bundle.

Maybe I should get the folder creation step to place a readme in legacy_folder directory to explain what it is for and not.

andrerom:
updated

emodric:
@andrerom If you think you can help, you can leave it as is. Since you don't have to use the command in your stack if empty folder structure bothers you, this should be okay :)

andrerom:
Ok, then who else should we get to review this then?

emodric:
> Ok, then who else should we get to review this then?

@wizhippo maybe ?

bundle/Command/LegacySrcSymlinkCommand.php
wizhippo: I'm with @emodric here on having `-c` as default.  Leaving off `-c` will throw the exception and help the users who have not read the docs be aware they can and should migrate their existing config to the new location.

andrerom:
> I too am curious how others are handling storage and should it be addressed, perhaps in another PR. I'm glad I have had backups a couple of times.


yes, at the very least would need to alert the user when we detect that folder being removed is not empty, like reported in #110.

bundle/Command/LegacySrcSymlinkCommand.php
andrerom: PR updated, also added much more output to try to explain the concepts better for user.

better? _(changes done in new commit)_

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -0,0 +1,159 @@
...
+            ->addOption('create', 'c', InputOption::VALUE_NONE, 'Create "src" directory structure if it does not exist')
+            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Symlink folders even if target already exist')
+            ->setDescription('Installs legacy project settings and design files from "src" to corresponding folders in ezpublish_legacy/')
+            ->setHelp(
+                <<<EOT
+The command <info>%command.name%</info> setups and symlinks <info>src/legacy_files</info> stored in your root project for
+any design/settings project files, and symlinks these into <info>ezpublish_legacy/</info> which is installed by composer.
+
+The benefit of this is:
+1. Being able to version your design/config files in git withouth versing legacy itself
```
glye: versing -> versioning

andrerom:
Thanks all for reviews!

Merging as I assume the only remaining concerns by @emodric and @wizhippo is now handled given structure is no longer automatically created, but instead instructions for how to do it is outputted.

PR #108, opened 2017-09-13
Upgraded PHPUnit to 5.7

MarioBlazek:
This PR resolves:
* tests now extend namespaced `TestCase`
* removed deprecated `getMock` method calls
* removed deprecated `setExpectedException` method calls
* removed Mockery library (we do not need it, just two stubs were created by it) -> replaced with PHPUnit syntax

Notice:
There are still some warning messages popping up, but those are coming from `ezpublish-kernel` .


emodric:
You can replace `$this->getMockBuilder(...)->disableOriginalConstructor()->getMock()` with `$this->createMock(...)`, but other than that, :+1:

emodric:
Thanks @MarioBlazek !

PR #111, opened 2017-09-30
Update doc about adding post-*-scripts

crevillo:
As we have now the build block which is referenced from `post-install-cmd` and `post-update-cmd` i though it would help to update the doc about this part. 

Let me know if i should create a ticket for this.


andrerom:
Thanks @crevillo, but similar thing going on here: https://github.com/ezsystems/LegacyBridge/pull/107/files

crevillo:
Oh, i missed that one. I think the other will fix more things in a row, so closing this to avoid noise. Thanks @andrerom

PR #112, opened 2017-10-03
legacy:assets_install should not delete non empty folders unless asked to

### LINKED ISSUE #110: ezpublish:legacy:assets_install should prompt user before deleting non empty folders
Issue: eZ Platform install with already existing image/binary files in `web/var`, folder with all files will be deleted by this command regardless of being empty or not when setting up symlink from `ezpublish_legacy/var`. 

So `ezpublish:legacy:assets_install` should when encountering non-empty folders, either:
- prompt user before removing
- move files over for the user and inform about it
- skip the folder and give warning to user that command needs to be re-executed once files have been moved over.
- ...

andrerom:
Fixes #110

### DIFF
--- a/bundle/Command/LegacySrcSymlinkCommand.php
+++ b/bundle/Command/LegacySrcSymlinkCommand.php
@@ -28,7 +28,7 @@ class LegacySrcSymlinkCommand extends ContainerAwareCommand
                 )
             )
             ->addOption('create', 'c', InputOption::VALUE_NONE, 'Create "src" directory structure if it does not exist')
-            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Symlink folders even if target already exist')
+            ->addOption('force', null, InputOption::VALUE_NONE, 'Force symlinking folders even if target already exist')
             ->setDescription('Installs legacy project settings and design files from "src" to corresponding folders in ezpublish_legacy/')
             ->setHelp(
                 <<<EOT
--- a/bundle/Command/LegacySrcSymlinkCommand.php
+++ b/bundle/Command/LegacySrcSymlinkCommand.php
@@ -28,7 +28,7 @@ class LegacySrcSymlinkCommand extends ContainerAwareCommand
                 )
             )
             ->addOption('create', 'c', InputOption::VALUE_NONE, 'Create "src" directory structure if it does not exist')
-            ->addOption('force', null, InputOption::VALUE_NONE, 'Force symlinking folders even if target already exist')
+            ->addOption('force', null, InputOption::VALUE_NONE, 'Force symlinking folders even if target already exists')
             ->setDescription('Installs legacy project settings and design files from "src" to corresponding folders in ezpublish_legacy/')
             ->setHelp(
                 <<<EOT
--- a/bundle/Command/LegacyWrapperInstallCommand.php
+++ b/bundle/Command/LegacyWrapperInstallCommand.php
@@ -66,12 +66,12 @@ EOT
             // Check if directory exists (not link) and is not empty to avoid removing things that should be backend up
             if (!$force && !is_link($targetDir) && is_dir($targetDir) && count(glob("$targetDir/*", GLOB_NOSORT)) > 0) {
                 $output->writeln(<<<EOT
-Skipping: The folder  "$targetDir" already exists and seems to contain content!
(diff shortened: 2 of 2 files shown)


bundle/Command/LegacyWrapperInstallCommand.php
```diff
@@ -56,10 +57,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
         foreach (array('design', 'extension', 'share', 'var') as $folder) {
             $targetDir = "$targetArg/$folder";
             $originDir = "$legacyRootDir/$folder";
+
+            // Check if directory exists (not link) and is not empty to avoid removing things that should be backend up
+            if (!$force && !is_link($targetDir) && is_dir($targetDir) && count(glob("$targetDir/*", GLOB_NOSORT)) > 0) {
+                $output->writeln(<<<EOT
+Skipping: The folder "$targetDir" already exists and seems to contain content!
+
+Make sure to backup this content and move it into corresponding legacy folder which is will be setup to symlink / copy
```
emodric: `which will be setup`

bundle/Command/LegacySrcSymlinkCommand.php
```diff
@@ -28,7 +28,7 @@ protected function configure()
                 )
             )
             ->addOption('create', 'c', InputOption::VALUE_NONE, 'Create "src" directory structure if it does not exist')
-            ->addOption('force', 'f', InputOption::VALUE_NONE, 'Symlink folders even if target already exist')
+            ->addOption('force', null, InputOption::VALUE_NONE, 'Force symlinking folders even if target already exist')
```
emodric: `if target already exists`

bundle/Command/LegacyWrapperInstallCommand.php
```diff
@@ -56,10 +57,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
 
         $output->writeln(sprintf("Installing eZ Publish legacy assets from $legacyRootDir using the <comment>%s</comment> option", $input->getOption('symlink') ? 'symlink' : 'hard copy'));
         $symlink = $input->getOption('symlink');
+        $force = (bool)$input->getOption('force');
 
         foreach (array('design', 'extension', 'share', 'var') as $folder) {
             $targetDir = "$targetArg/$folder";
             $originDir = "$legacyRootDir/$folder";
+
+            // Check if directory exists (not link) and is not empty to avoid removing things that should be backend up
```
glye: backed up

bundle/Command/LegacyWrapperInstallCommand.php
```diff
@@ -56,10 +57,28 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+            if (!$force && !is_link($targetDir) && is_dir($targetDir) && count(glob("$targetDir/*", GLOB_NOSORT)) > 0) {
+                $output->writeln(<<<EOT
+Skipping: The folder "$targetDir" already exists and seems to contain content!
+
+Make sure to backup this content and move it into corresponding legacy folder which is will be setup to symlink / copy
+to this folder before you remove it, then re-run this command.
+
+If this folder and the other "$targetArg" directories can be safely overwritten, run this command with <info>--force</info> option.
```
glye: command with **the**

andrerom:
> Assuming you have checked that it doesn't count . and .. as folder content :)

from comments and doc it does not include those, it also does not include hidden folders _(`.*`)_which is fine in this case.

andrerom:
Thanks for the reviews, sorry if the only things to comment on is grammar as always :P

blankse:
I have a problem with this:
The flysystem Local adapter creates a web/var/storage folder.
The composer ScriptHandler creates no symlink for web/var anymore because the storage folder counts as content.

andrerom:
@blankse So you are referring to setup that has already been used running platform, and then you add legacy bridge? or during install of both?

blankse:
During install of both. If the symlink already exists everything works fine.
But if you build it new from stratch (e. g. platform.sh) the symlink doesn't exist and the Local Adapter (Dependency from IOBinarydataHandler/IOMetadatahandler) create a web/var/storage folder. This folder blocks the creating of the symlink to ezpublish_legacy/var folder.

PR #116, opened 2017-10-10
Add missing siteaccess option definition to LegacyEmbedScriptCommand

blankse:
In line 63 the option is used but is not defined.

If you want use it, an error occurs.


blankse:
Oh, the `$this->ignoreValidationErrors();` should avoid the error.

But why it doesn't happens in my case i don't no. I dig deeper.

andrerom:
Afaik `siteaccess` is defined globally by eZ Platform, but long time since I looked into that part so don't remember how it is done.

blankse:
Yes, eZ Platform has a own Application class (https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Console/Application.php), which set the siteaccess option. In my case I use the symfony one.

We can close this PR.

PR #117, opened 2017-10-16
EZP-28084: Update ScriptHandler to support symfony-bin-dir

wizhippo:
Add support for Symfony 3.0 directory structure

https://jira.ez.no/browse/EZP-28084

### DIFF
--- a/bundle/Composer/ScriptHandler.php
+++ b/bundle/Composer/ScriptHandler.php
@@ -29,7 +29,7 @@ class ScriptHandler extends DistributionBundleScriptHandler
     public static function installAssets(Event $event)
     {
         $options = self::getOptions($event);
-        $appDir = $options['symfony-app-dir'];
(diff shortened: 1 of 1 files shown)


andrerom:
Good catch :)

Note that there should be a method already available for this. See kernel for approach there: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Composer/ScriptHandler.php#L47

emodric:
>Note that there should be a method already available for this. See kernel for approach there: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Composer/ScriptHandler.php#L47

Was about to comment the same thing :)

wizhippo:
Updated to use existing `getConsoleDir`.

Thanks for pointing it out @andrerom @emodric

emodric:
Thanks @wizhippo !

PR #119, opened 2017-10-24
Suggested edits to clarify the need for setup wizard

andrerom:


emodric:
Looks good!

As for legacy field types, there's https://github.com/netgen/NetgenRichTextDataTypeBundle which is available for a year or so, which can be mentioned in the docs :)

andrerom:
Nice, I was not aware of this! :)

andrerom:
@emodric better?

PR #121, opened 2017-10-31
Fix HttpCachePass after #120

emodric:


andrerom:
+1, but I guess we could have reused definition from the one that was removed or? (to avoid duplicating too much definitions in compiler passes)

emodric:
You mean something like this?

`$container->setAlias('ezplatform.http_cache.purger', 'ezpublish.http_cache.purger.instant');`

andrerom:
ok, let's go with this to unblock :)

PR #122, opened 2017-11-21
Support for eZ kernel 7.0

emodric:
Just an unlock in `composer.json` for now, but obviously will need some more work :)


andrerom:
Any idea what else will be needed? Asking as it might not need a new major version unless it involves breaks.

emodric:
At the very least, support for the new HTTP cache bundle. I think that it's a breaking change, but not sure.

bundle/DependencyInjection/Compiler/HttpCachePass.php
```diff
@@ -31,6 +31,7 @@ public function process(ContainerBuilder $container)
                     new Reference('ezplatform.http_cache.purge_client'),
                     new Reference('ezpublish.api.service.inner_content'),
                     new Reference('ezpublish.http_cache.event_dispatcher'),
+                    new Reference('ezpublish.api.repository'),
```
andrerom: this is also relevant for upcoming 6.7.7, 6.12.1 and 6.13.0.
Can code be done in different way to avoid having to set all arguments?

andrerom: https://github.com/ezsystems/ezpublish-kernel/commit/e778d2bd6ee2ba850f420f1fb729b832d8b6c342#diff-756d49898b2a4da2e9d185f1de4f4248

emodric: I'll try to find a way. This is just a playground for now :)

emodric: I don't see a way to fix this any other way :/

The problem is that by this point any reference to old `ezpublish.http_cache.purger.instant` is removed from the container and we simply do not know the list of arguments.

bundle/Cache/PersistenceCachePurger.php
```diff
@@ -69,7 +69,7 @@ public function all()
             return;
         }
 
-        $this->cache->clear();
+        //$this->cache->clear();
```
emodric: @andrerom I'm going to need some help with this one, or rather, with any commented out call to cache client here...

andrerom: Indeed.

well first we need to decide if we are going to have a branch only supporting 7.0 or have two adapters here being injected depending on version.

emodric: I'm all for keeping it simple. 2.0 legacy bridge supporting only kernel 7.0.

If needed, this adapter could be backported than to 1.x.

andrerom: Ok, I'll push commit for this one.

emodric: Don't forget other commented out calls, those are all I didn't know what to do with :)

andrerom:
@emodric pushed, not tested yet here, but on theoretical side one relevant todo left there in regards to clearing type cache on group changes. The user one is most likely not needed given how it is used.

emodric:
@andrerom Thanks a lot! It looks okay with only a reading of the code. We can always iterate later if something is missing :)

emodric:
@andrerom In this state it is okay with me to merge. WDYT? I was thinking of removing `*.class` parameters too, and marking services as private, but I can do it in separate PR.

emodric:
So far so good, no issues.

PR #123, opened 2017-11-23
Add missing argument to ezpublish.http_cache.purger.instant

emodric:
For support for 6.7.7, 6.12.1 and 6.13.0

Still there is no way to properly solve this since at this point any reference to old `ezpublish.http_cache.purger.instant` is removed from the container and we simply do not know the list of arguments.


emodric:
@andrerom Ready for merge and tag?

andrerom:
https://github.com/ezsystems/LegacyBridge/releases/tag/v1.4.1

emodric:
:+1:

PR #125, opened 2017-12-11
legacy:assets_install should remove empty IO root before check empty var folder

blankse:
See comments here: https://github.com/ezsystems/LegacyBridge/pull/112


blankse:
@andrerom Here a solution which works for me. But if someone set a custom var dir (e. g. `var/site` like in the default ezplatform.yml) I think this don't work.

Perhaps we can check empty folders with scandir?
`array_diff(scandir($directory), array('..', '.'))`

So empty sub directories make no problems.

What do you think?

PR #126, opened 2017-12-16
Apply Symfony best practices

emodric:
* Remove class parameters
* use `kernel.project_dir` instead of `kernel.root_dir`
* Use Twig namespaces for templates
* Marked all services as private/public

### DIFF
--- a/bundle/Controller/LegacySetupController.php
+++ b/bundle/Controller/LegacySetupController.php
@@ -8,20 +8,20 @@
  */
 namespace eZ\Bundle\EzPublishLegacyBundle\Controller;
 
+use eZ\Bundle\EzPublishLegacyBundle\SetupWizard\ConfigurationConverter;
+use eZ\Bundle\EzPublishLegacyBundle\SetupWizard\ConfigurationDumper;
 use eZ\Publish\Core\MVC\Legacy\Kernel\Loader;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
 use eZ\Publish\Core\MVC\Symfony\ConfigDumperInterface;
 use eZ\Bundle\EzPublishLegacyBundle\DependencyInjection\Configuration\LegacyConfigResolver;
 use eZ\Bundle\EzPublishLegacyBundle\Cache\PersistenceCachePurger;
 use eZINI;
 use eZCache;
+use Symfony\Component\HttpKernel\KernelInterface;
 
-class LegacySetupController implements ContainerAwareInterface
+class LegacySetupController
 {
-    use ContainerAwareTrait;
-
     /**
      * The legacy kernel instance (eZ Publish 4).
      *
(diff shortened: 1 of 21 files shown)


andrerom:
ping @lolautruche

bundle/Controller/LegacySetupController.php
```diff
@@ -42,31 +42,58 @@ class LegacySetupController implements ContainerAwareInterface
...
 
     /**
      * @param \Closure $kernelClosure
      * @param \eZ\Bundle\EzPublishLegacyBundle\DependencyInjection\Configuration\LegacyConfigResolver $legacyConfigResolver
      * @param \eZ\Bundle\EzPublishLegacyBundle\Cache\PersistenceCachePurger $persistenceCachePurger
      * @param \eZ\Publish\Core\MVC\Legacy\Kernel\Loader $kernelFactory
+     * @param \eZ\Bundle\EzPublishLegacyBundle\SetupWizard\ConfigurationDumper $configDumper
+     * @param \eZ\Bundle\EzPublishLegacyBundle\SetupWizard\ConfigurationConverter $configConverter
+     * @param \Symfony\Component\HttpFoundation\RequestStack $requestStack
+     * @param \Symfony\Component\HttpKernel\KernelInterface
```
crevillo: Missed var name in this line?

lolautruche: I think the whole PHPDoc could be removed

bundle/Resources/config/default_settings.yml
```diff
@@ -5,7 +5,7 @@ parameters:
     ezpublish_legacy.default.module_default_layout: ~
 
     # Pagelayout to use while rendering a content view in legacy
-    ezpublish_legacy.default.view_default_layout: EzPublishLegacyBundle::legacy_view_default_pagelayout.html.twig
+    ezpublish_legacy.default.view_default_layout: "@@EzPublishLegacy/legacy_view_default_pagelayout.html.twig"
```
lolautruche: Why a double `@` here?

bundle/Controller/LegacySetupController.php
emodric: Yep, thanks. Will fix it tomorrow, when I'm at my laptop :)

bundle/Resources/config/default_settings.yml
emodric: Without it, Symfony crashes with unknown service error or something similar. I can't find the doc about it ATM, but I have it working like this on some of our other projects, after I've read it in the docs 🙊

emodric:
@lolautruche I'll write up some 2.0 upgrade docs, thanks 👍

bundle/Resources/config/default_settings.yml
emodric: Ha, found it! 

http://symfony.com/doc/current/service_container/parameters.html

> If you use a string that starts with @ or has % anywhere in it, you need to escape it by adding another `@` or `%`

```
# config/services.yaml
parameters:
    # This will be parsed as string '@securepass'
    mailer_password: '@@securepass'

    # Parsed as http://symfony.com/?foo=%s&amp;bar=%d
    url_pattern: 'http://symfony.com/?foo=%%s&amp;bar=%%d'
```

emodric:
Note to self: Need to rebase on master to get  the tests running once https://github.com/ezsystems/LegacyBridge/pull/122 is merged.

andrerom:
> Note to self: Need to rebase on master to get the tests running once #122 is merged.

merged

mvc/Kernel/Loader.php
```diff
@@ -263,7 +264,7 @@ public function setCLIHandler(ezpKernelHandler $kernelHandler)
     public function buildLegacyKernelHandlerTreeMenu()
     {
         return $this->buildLegacyKernelHandlerWeb(
-            $this->container->getParameter('ezpublish_legacy.kernel_handler.treemenu.class'),
+            ezpKernelWeb::class,
```
blankse: Shouldn't it be ezpKernelTreeMenu?

emodric: Eeeek, you're right!!

Fixed, thanks!

emodric:
@andrerom @crevillo @blankse @lolautruche Ready for final review and merge :)

Will writeup upgrade docs in a separate PR.

emodric:
Thanks everyone!

bundle/Resources/config/templating.yml
```diff
@@ -26,53 +17,64 @@ services:
...
+        class: eZ\Publish\Core\MVC\Legacy\Templating\GlobalHelper
         parent: ezpublish.templating.global_helper.core
+        public: false
         calls:
             - [setLegacyHelper, ["@ezpublish_legacy.templating.legacy_helper"]]
 
     # Overriding core helper
     ezpublish.templating.global_helper:
         alias: ezpublish.templating.global_helper.legacy
+        public: false
```
wizhippo: Is EzPublishCoreBundle\Controller deprecated?   The breaks it's 'getGlobalHelper' as it needs the service to be public or an exception is thrown and it is private now.

andrerom: My guess: helper should be adapted to inject the service instead of relying on contianer.

wizhippo: I don't believe when aliasing a service we should change its viability either which is causing me the issue I see.

emodric:
I'm fine with it being public, no problem :)

bundle/Resources/config/templating.yml
emodric: I'm fine with it being public, no problem :)

lolautruche: Another approach would be to use the [service locator feature](https://symfony.com/blog/new-in-symfony-3-3-service-locators) in `EzPublishCoreBundle\Controller` (Symfony 3.3+)

lolautruche: See implementation in Symfony `AbstractController`: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php

xserna: Hi guys, I've found this issue after upgrading from an old 5.4 to the new 2.5.6, solved by overriding service definition on my own services.yml file.

andrerom: Someone willing to do PR to drop this, I agree we should not change this from kernel.

xserna: PR: https://github.com/ezsystems/LegacyBridge/pull/175

PR #127, opened 2017-12-18
Remove deprecated features

emodric:
* Removes Twig tag rendering legacy templates
* Removes configuring if HTTP 404 exception conversion will happen or not
* Replaces `DefinitionDecorator` with `ChildDefinition` as per deprecation notice


andrerom:
Which of these breaks will affect people upgrading from 5.4?

emodric:
Legacy Twig template rendering tag was deprecated in 5.1, so I hope that one would be safe. But it should be pretty straightforward to fix with a search and replace.

Not doing 404 exception conversion was always a bug in my eyes, rather than a feature. Having to implement error pages in legacy sucked big time IMO. But I guess it is a break if someone still has error pages in legacy.

Third one should not matter.

I will write up the upgrade docs, yes, as soon as we finalize 2.0.

andrerom:
> Not doing 404 exception conversion was always a bug in my eyes, rather than a feature. Having to implement error pages in legacy sucked big time IMO. But I guess it is a break if someone still has error pages in legacy.

with this change it now depends on legacy mode right? so only legacy handles 404 when legacy mode is on? _If so, much more clear for me to at least, so +1_

_Thinking a bit if we (eZ, @ezsystems/ez-support-team ) should double check if we should backport this deprecation to 5.4 then so customers gets a advance warning about this._

> Third one should not matter.

yes this one is +1

> Legacy Twig template rendering tag was deprecated in 5.1, so I hope that one would be safe. But it should be pretty straightforward to fix with a search and replace.

This one might be a problem if both we and customers still use it, or was this replaced by another mechanism instead? /cc @lolautruche / @bdunogier

emodric:
> with this change it now depends on legacy mode right? so only legacy handles 404 when legacy mode is on? If so, much more clear, so +1

Yep.

> This one might be a problem if both we and customers still use it, or was this replaced by another mechanism instead? /cc @lolautruche / @bdunogier

It now uses native Twig `include` tag, since legacy template engine is implemented as a real templating engine in Symfony. As I saif, migration is pretty much straightforward  search and replacing one tag with the other.

andrerom:
> It now uses native Twig include tag, since legacy template engine is implemented as a real templating engine in Symfony. As I saif, migration is pretty much straightforward search and replacing one tag with the other.

ok, we definitely need upgrade doc then ;)

emodric:
@andrerom So, ready for merge? Anything I need to do for backport of 404 conversion to 5.4?

andrerom:
> Anything I need to do for backport of 404 conversion to 5.4?

nope

anyone up for second review @lolautruche / @crevillo

emodric:
Thanks everyone!

PR #128, opened 2017-12-19
Add 2.0 upgrade docs

emodric:
Documents upgrading to 2.0, specifically:

https://github.com/ezsystems/LegacyBridge/pull/122
https://github.com/ezsystems/LegacyBridge/pull/126
https://github.com/ezsystems/LegacyBridge/pull/127

### DIFF
--- a/doc/upgrade/2.0.md
+++ b/doc/upgrade/2.0.md
@@ -1,0 +1,37 @@
+# UPGRADE FROM 1.x to 2.0
+
+* Support for eZ Publish kernel 6.x has been dropped. Minimum supported version
+  is now 7.0
+
+* Support for PHP 5.x and 7.0 has been dropped. Minimum supported version is
+  now 7.1
+
+* All `*.class` parameters were removed from DI container. Use compiler passes
+  to set the new class or service decorators if you need to override any
+  services from legacy bridge.
+  
+* Most of the services have been marked as private in Symfony DI container.
+  You should not fetch them from container directly anymore, instead inject
+  them as constructor arguments into your services. 
+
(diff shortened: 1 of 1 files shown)


doc/upgrade/2.0.md
```diff
@@ -0,0 +1,36 @@
...
+
+* Support for PHP 5.x and 7.0 has been dropped. Minimum supported version is
+  now 7.1
+
+* All `*.class` parameters were removed from DI container. Use compiler passes
+  to set the new class or service decorators if you need to override any
+  services from legacy bridge.
+  
+* Most of the services have been marked as private in Symfony DI container.
+  You should not fetch them from container directly any more, instead, inject
```
andrerom: anymore?

doc/upgrade/2.0.md
```diff
@@ -0,0 +1,36 @@
...
+  them as constructor arguments into your services. 
+
+* HTTP 404 error conversion in legacy mode `false` siteaccesses is now turned
```
andrerom: something like this might be easier to grasp:`HTTP 404 error conversion when legacy mode is `false` in given siteaccesses is now always handled by Symfony, and config option to control this has been removed so the logic follows legacy mode setting.`

emodric:
@andrerom Fixed both nitpicks.

andrerom:
And this one made the other PR much more clear _(which is why dealing with doc in same PR as code is sometimes a good thing ;))_

emodric:
> (which is why dealing with dev doc in same PR as code is sometimes a good thing ;))

;)

PR #129, opened 2017-12-19
Require eZ Platform XMLText field type

emodric:


andrerom:
Don't forget to update doc @emodric ;)

emodric:
Right, will do :)

PR #134, opened 2018-01-03
Fix incompatibility with Http Cache Bundle >=0.4

emodric:
Basically loads a different version of SwitchableHttpCachePurger based on availability of eZ Platform HTTP cache bundle.


emodric:
@andrerom When merging to `master`, all of these changes should be discarded ofcourse.

PR #135, opened 2018-01-10
[Feature] Set legacy in DevMode & DebugOuput when symfony is in debug

andrerom:
Could have checked for kernel.env === dev also in some form, but seems seemed cleaner, up for debate if we want this tough.

### DIFF
--- a/bundle/LegacyMapper/Configuration.php
+++ b/bundle/LegacyMapper/Configuration.php
@@ -180,7 +180,7 @@ class Configuration implements EventSubscriberInterface
         // Cache settings
         // Enforce ViewCaching to be enabled in order to persistence/http cache to be purged correctly.
         $settings['site.ini/ContentSettings/ViewCaching'] = 'enabled';
-        
+
         // Dev / Debug
         // if symfony is in debug mode, also put legacy in dev mode as a convention for ease of use
         if ($this->container->getParameter('kernel.debug')) {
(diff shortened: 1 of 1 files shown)


emodric:
@andrerom What are the usecases? I'm not a fan of this change, especially since it cannot be configured/disabled. Nowadays, legacy development, at least at Netgen, is pretty much nonexistent, as we run only legacy admin siteaccesses in eZ legacy.

andrerom:
True, so you don't want DebugOuput (maybe unless we route that to symfony toolbar).
But anyway, first stage, DevelopmentMode could make sense for everyone right? Only thing it does is add checks to auto refresh templates cache and view cache when relevant templates change.

emodric:
> But anyway, first stage, DevelopmentMode could make sense for everyone right? Only thing it does is add checks to auto refresh templates cache and view cache when relevant templates change.

Which again slows down everything :) Plus, it disables eZ JS Core concatenating JS and CSS assets. So I personally wouldn't want any of those two enabled. I'd be fine with it if it were behind a flag (initially disabled).

andrerom:
yeah. I understand you. Any legacy developer who knows all this will agree with you. Persona I was optimizing for was someone unaware of how legacy works and who sometimes needs to take a look, hence just installs legacy to check how things functioned there, and who would benefit from getting debug / dev stuff to auto enable if symfony was in dev mode.

If there was a way to set this at a lower ini priority level, so any defined siteaccess or override setting would take precedent, then that might be a good mix. But as long as that is not that simple, I would also not optimize for a couple of users on expense of 99% of of legacy users.

emodric:
Maybe something like this could be used to half-automate the config: https://github.com/fagbokforlaget/ForlagshusetLegacyIniBundle (as in, activate a custom bundle which would automatically enable the config).

This was originally written by me and extracted by our clients to reuse outside of context of Netgen projects.

dfritschy:
Fully agree with not merging this PR. Nice solution by Edi!

andrerom:
In related news: Has anyone done work on routing debug output to symfony web toolbar? /cc @lolautruche ?

PR #138, opened 2018-01-12
Only symlink top-level directories in LegacySrcSymlinkCommand

natanael89:
In LegacySrcSymlinkCommand, only the top-level directories need to be symlinked. The rest is redundant.


andrerom:
And then I realize both PR's should have been towards 1.4 branch ☠️ 
I'll cherry-pick and merge then.. 🍳

PR #139, opened 2018-01-15
Enabled PHPUnit v6

MarioBlazek:


emodric:
@mblazek Can you use the caret operator `^` in `composer.json` instead of tilda `~`?

emodric (approved):

andrerom (approved):

PR #140, opened 2018-01-22
EZP-28736: SPI cache not cleared when a content is published via a delayed publication workflow

adamwojs:
> JIRA: https://jira.ez.no/browse/EZP-28736

## Description

This PR adds the second optional argument to `\eZ\Bundle\EzPublishLegacyBundle\Cache\PersistenceCachePurger::content` for affected objectIds, and use it for clearing the relevant content cache. This allows us to avoid of location loading in cache purging, which may cause issues when cache purging is executed in the in the middle of a transaction started by the legacy database connection.   

Changes suggested in https://github.com/ezsystems/ezpublish-legacy/pull/1343#issuecomment-357519202

- [x] Adjust the `ezcontentcachemanager::clearNodeViewCacheArray` / `eZContentOperationCollection::removeNodes` calls in ezpublish-legacy (https://github.com/ezsystems/ezpublish-legacy/pull/1345)

### DIFF
--- a/bundle/Cache/PersistenceCachePurger.php
+++ b/bundle/Cache/PersistenceCachePurger.php
@@ -101,13 +101,13 @@ class PersistenceCachePurger implements CacheClearerInterface
      * Either way all location and urlAlias cache is cleared as well.
      *
      * @param int|int[]|null $locationIds Ids of location we need to purge content cache for. Purges all content cache if null
-     * @param int|int[]|null $contentIds Ids of content we need to purge
+     * @param int[]|null $contentIds Ids of content we need to purge
      *
      * @return array|int|\int[]|null
      *
      * @throws \eZ\Publish\Core\Base\Exceptions\InvalidArgumentType On invalid $id type
      */
-    public function content($locationIds = null, $contentIds = null)
+    public function content($locationIds = null, array $contentIds = null)
     {
         if ($this->allCleared === true || $this->isSwitchedOff()) {
             return $locationIds;
@@ -122,8 +122,6 @@ class PersistenceCachePurger implements CacheClearerInterface
 
         if ($contentIds === null) {
             $contentIds = array();
-        } elseif (!is_array($contentIds)) {
-            $contentIds = array($contentIds);
         }
 
         foreach ($locationIds as $id) {
@@ -139,6 +137,7 @@ class PersistenceCachePurger implements CacheClearerInterface
                 );
             }
         }
+
         foreach (array_unique($contentIds) as $id) {
             if (!is_scalar($id)) {
                 throw new InvalidArgumentType('$id', 'int[]|null', $id);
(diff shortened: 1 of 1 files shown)


bundle/Cache/PersistenceCachePurger.php
```diff
@@ -119,25 +120,37 @@ public function content($locationIds = null)
...
-                $this->cache->clear('user', 'role', 'assignments', 'byGroup', $location->contentId);
-                $this->cache->clear('user', 'role', 'assignments', 'byGroup', 'inherited', $location->contentId);
+                $contentIds[] = $this->locationHandler->load($id)->contentId;
             } catch (NotFoundException $e) {
                 $this->logger->notice(
                     "Unable to load the location with the id '$id' to clear its cache"
                 );
             }
         }
+        foreach ($contentIds as $id) {
```
andrerom: Make this unique

bundle/Cache/PersistenceCachePurger.php
```diff
@@ -119,25 +120,37 @@ public function content($locationIds = null)
...
+            if (!is_scalar($id)) {
+                throw new InvalidArgumentType('$id', 'int[]|null', $id);
+            }
+
+            $this->cache->clear('content', $id);
+            $this->cache->clear('content', 'info', $id);
+            $this->cache->clear('content', 'info', 'remoteId');
```
andrerom: can move this down into related cache while at it, does not need to be in the loop

crevillo: and what about move these 6 calls to a dedicated method? (maybe protected)

andrerom: nah, they are not relevant for master

adamwojs:
PR updated according to @andrerom suggestions.

bundle/Cache/PersistenceCachePurger.php
```diff
@@ -101,12 +101,13 @@ public function resetAllCleared()
      * Either way all location and urlAlias cache is cleared as well.
      *
      * @param int|int[]|null $locationIds Ids of location we need to purge content cache for. Purges all content cache if null
+     * @param int|int[]|null $contentIds Ids of content we need to purge
```
andrerom: could we type hint this as int[]|null? So we can add "array" on method signature? Pr do you see any specific needs to support both?

bundle/Cache/PersistenceCachePurger.php
```diff
@@ -119,28 +120,40 @@ public function content($locationIds = null)
...
-                $this->cache->clear('user', 'role', 'assignments', 'byGroup', $location->contentId);
-                $this->cache->clear('user', 'role', 'assignments', 'byGroup', 'inherited', $location->contentId);
+                $contentIds[] = $this->locationHandler->load($id)->contentId;
             } catch (NotFoundException $e) {
                 $this->logger->notice(
                     "Unable to load the location with the id '$id' to clear its cache"
                 );
             }
         }
+        foreach (array_unique($contentIds) as $id) {
```
andrerom: empty line between the two loops

PR #144, opened 2018-03-03
remove the warning of obsolete use without a single quotation mark

julienlegrand07:
FIX : https://jira.ez.no/browse/EZP-28891


emodric:
Thanks :)

PR #146, opened 2018-03-22
EZP-28926: Add a legacy security voter decorator

emodric:
This allows checking for legacy modules/functions before exceptions are thrown.

It only rethrows the exception if for some reason this voter votes with `ACCESS_ABSTAIN`.

Needed for legacy permission checking after https://github.com/ezsystems/ezpublish-kernel/pull/2273.

### DIFF
--- a/mvc/Security/Voter/VoterDecorator.php
+++ b/mvc/Security/Voter/VoterDecorator.php
@@ -1,0 +1,82 @@
+<?php
+/**
+ * This file is part of the eZ LegacyBridge package.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace eZ\Publish\Core\MVC\Legacy\Security\Voter;
+
+use Closure;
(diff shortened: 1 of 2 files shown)


andrerom:
> Needed for legacy permission checking after

You mean netgen legacy admin or something? legacy extensions tend to use legacy api for this right? so not really affected or?

emodric:
Meaning any custom code from the new stack checking for legacy permissions.

We have some pages implemented in the frontend too which reuse pieces of legacy, while checking for legacy permissions too.

mvc/Security/Voter/VoterDecorator.php
```diff
@@ -0,0 +1,82 @@
...
+    {
+        return $attribute instanceof AuthorizationAttribute;
+    }
+
+    public function supportsClass($class)
+    {
+        return true;
+    }
+
+    public function vote(TokenInterface $token, $object, array $attributes)
```
kmadejski: Would be nice to see PHPDocs here.

emodric:
@kmadejski Added PHPDoc :+1:

emodric:
@andrerom OK to merge and tag?

PR #147, opened 2018-04-10
[FEATURE]  Add command "init" to configure platform install for legacy bridge usage

andrerom:
In order to  simplify upgrades from pure legacy installs, this command is meant to be used like this to simplify install / upgrades _(in comparison to todays INSTALL.md, which does not cover upgrades, and is a bit more manual still)_:
- Enable bundle(s)
- (Optional) Edit config for legacy routes, & enable legacy mode for backend
- Install legacy bridge
- Run `init` which will:
  - [x] Generate all legacy related folders, and tell you which one is which and where to copy your files from eZ Publish
  - Adjust composer.json as current described in INSTALL doc automatically ?
    - [x] Add the composer legacy post-*-scripts
    - [x]  Enable EzPublishLegacyBundle
    - [x] Add legacy routes
    - ~Optional: Enable legacy_mode for your backoffice siteaccess~
    - ~Optional: add security rules for the Setup Wizard~
    - ~Optional: Ask user to install recommended extensions / bundles (search ext, richtext ext, ..)~
    - [x] Instruct user to run `composer run-script post-update-cmd` once things have been moved into folders, **or maybe rather `composer legacy-scripts`**
-  [x] Update INSTALL doc to reflect usage of this command.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -3,76 +3,40 @@
 Instructions below will take you true installing legacy bridge and implicit legacy on top of a eZ Platform 1.13.x - 2.x
 install.
 
-### Add the composer `legacy post-*-scripts`
+_TIP:_
+> Before starting make sure to check-in (e.g. to Git) your eZ Platform project working space so you'll be able to see & verify changes applied to your setup separate from initial clean project install.
 
-Edit `composer.json`, and add those lines to both `post-update-cmd` and `post-install-cmd` blocks at the end, but before
-`eZ\Bundle\EzPublishCoreBundle\Composer\ScriptHandler::dumpAssets`:
+### Install `ezsystems/legacy-bridge` and run `init` command
+
+1. Installed package using [Composer](https://getcomposer.org/doc/00-intro.md):
 ```
-"scripts": {
-    "legacy-scripts": [
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installAssets",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacyFiles"
-    ],
-    "post-install-cmd": [
-        ...,
-        "@legacy-scripts"
-    ],
-    "post-update-cmd": [
-        ...,
-        "@legacy-scripts"
-    ],
-}
+composer require --update-no-dev "ezsystems/legacy-bridge"
 ```
 
-Example: In the case of stock eZ Platform 1.13 that would specifically be:
-```
-"scripts": {
-    "legacy-scripts": [
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installAssets",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads",
-        "eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacyFiles"
-    ],
-    "symfony-scripts": [
-        "Incenteev\\ParameterHandler\\ScriptHandler::buildParameters",
-        "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::buildBootstrap",
-        "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::clearCache",
-        "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installAssets",
-        "@legacy-scripts",
-        "eZ\\Bundle\\EzPublishCoreBundle\\Composer\\ScriptHandler::dumpAssets",
-        "Sensio\\Bundle\\DistributionBundle\\Composer\\ScriptHandler::installRequirementsFile"
-    ],
-    "post-install-cmd": [
-        "@symfony-scripts"
-    ],
-    "post-update-cmd": [
-        "@symfony-scripts"
-    ],
-    (...)
-```
-
-
-### Enable EzPublishLegacyBundle
+2. Enable `EzPublishLegacyBundle`
 Edit `app/AppKernel.php`, and add `new eZ\Bundle\EzPublishLegacyBundle\EzPublishLegacyBundle( $this ),`
-at the end of  the `$bundles` array. Pay close attention to the `$this` argument, LegacyBundle is a bit 
-spoiled and has high expectations from its collaborators ;)
+at the end of the `$bundles` array _(typically just after `new AppBundle\AppBundle(),`)_.
 
-### Add legacy routes
-Edit `app/config/routing.yml`, and add the LegacyBundle routes at the end of the file.
+_NOTE: Pay close attention to the `$this` argument, LegacyBundle needs it to interact with other eZ bundles._
 
+3. Run the following command to configure your install for legacy usage:
 ```
-# NOTE: Always keep at the end of the file so native symfony routes always have precendence, to avoid legacy
-# REST pattern overriding possible eZ Platform REST routes. 
-_ezpublishLegacyRoutes:
-    resource: @EzPublishLegacyBundle/Resources/config/routing.yml
+php app/console ezpublish:legacy:init
 ```
+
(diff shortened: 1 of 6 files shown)


emodric:
@andrerom Are you sure this is neeeded? Isn't symlinking of legacy extensions already handled from any bundle that has a `ezpublish_legacy` folder? What I was talking about (issues with extension ordering) was when you use such bundles to also automatically activate the extensions, which is not a requirement for symlinking to work.

andrerom:
Hmm.. Then maybe not, there could be conflicts if developer is doing something funky. On the other hand this is far more straight forward if you are upgrading from 4.x, no need for a bunlde which might not be relevant once you have moved all code to platform stack.

emodric:
> On the other hand this is far more straight forward if you are upgrading from 4.x, no need for a bundle which might not be relevant once you have moved all code to platform stack.

True, but then it needs to be documented well that this incompatible with `eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions` script which symlinks legacy extensions from the bundles.

andrerom:
@emodric Different approach, haven't really tested it yet, just want to check if direction makes sense for you. Goal here is the same, make it more clear for people migrating where to place files coming form 4.x or 5.x.

andrerom:
@emodric what do you think?

ref:
bin/console ezpublish:legacy:init --help
> "Inits folders you can use when migrating from eZ Publish install to eZ Platform with Legacy Bridge"

bundle/Command/LegacyInitCommand.php
```diff
@@ -0,0 +1,88 @@
...
+                array(
+                    new InputArgument('src', InputArgument::OPTIONAL, 'The src directory for legacy files', 'src/legacy_files'),
+                )
+            )
+            ->setDescription('Inits folders you can use when migrating from eZ Publish install to eZ Platform with Legacy Bridge')
+            ->setHelp(
+                <<<EOT
+The command <info>%command.name%</info> creates the following folders for migration usage:
+- src/AppBundle/ezpublish_legacy => Optionaly for extensions you want to version in your project source (as opposed to install using composer)
+- src/legacy_files/design  => Optionally for cusotm designs not already made part of an extension
```
emodric: custom

andrerom:
@emodric PR description updated to reflect better the overall workflow I'm suggesting here.

This might be something @vidarl, @crevillo, @joekepley and @peterkeung might want to give input on as well here.

vidarl:
Implemented "Adjust composer.json as current described in INSTALL doc automatically"...

bundle/Command/LegacyInitCommand.php
```diff
@@ -90,4 +91,46 @@ protected function createDirectories(InputInterface $input, OutputInterface $out
...
+            $composerJson['scripts'] = array_merge(['legacy-scripts' => $legacy_scripts], $composerJson['scripts']);
+            $updateComposerJson = true;
+        }
+
+        if (!array_key_exists('symfony-scripts', $composerJson['scripts'])) {
+            $composerJson['scripts']['symfony-scripts'] = ['@legacy-scripts'];
+            $updateComposerJson = true;
+        } elseif (!in_array('@legacy-scripts', $composerJson['scripts']['symfony-scripts'])) {
+            $symfonyScripts = $composerJson['scripts']['symfony-scripts'];
+            $offset = array_search('@php bin/console assetic:dump', $symfonyScripts);
```
andrerom: it has not always used dump command: https://github.com/ezsystems/ezplatform/blob/1.13/composer.json#L76

bundle/Command/LegacyInitCommand.php
```diff
@@ -0,0 +1,138 @@
...
+- src/AppBundle/ezpublish_legacy  _(for extensions)_
+- src/legacy_files/design
+- src/legacy_files/settings/override
+- src/legacy_files/settings/siteaccess
+
+Move over files and directories from older installation and afterwards run the folling commands to setup symlinks into
+legacy install:
+- <info>ezpublish:legacy:symlink</info>
+- <info>ezpublish:legacybundles:install_extensions</info>
+- <info>ezpublish:legacy:assets_install</info>
```
andrerom: these can now be changed to `composer legacy-scripts`

bundle/Command/LegacyInitCommand.php
```diff
@@ -0,0 +1,138 @@
...
+                'eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::installLegacyBundlesExtensions',
+                'eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::generateAutoloads',
+                'eZ\\Bundle\\EzPublishLegacyBundle\\Composer\\ScriptHandler::symlinkLegacyFiles',
+            ];
+            $composerJson['scripts'] = array_merge(['legacy-scripts' => $legacy_scripts], $composerJson['scripts']);
+            $updateComposerJson = true;
+        }
+
+        if (!array_key_exists('symfony-scripts', $composerJson['scripts'])) {
+            $composerJson['scripts']['symfony-scripts'] = ['@legacy-scripts'];
```
andrerom: In this case we should probably either write warning saying symfony scripts should be added post-install and post-update, or we should try to do that here(?)

andrerom:
@emodric / @vidarl Should be ready for review now, this can be expanded later with asking user to install recommended bundles / extensions as well.

bundle/Command/LegacyConfigurationCommand.php
```diff
@@ -59,5 +59,6 @@ protected function execute(InputInterface $input, OutputInterface $output)
         $configurationDumper->dump($configurationConverter->fromLegacy($package, $adminSiteaccess), $options);
 
         $output->writeln('Configuration written to ezpublish.yml and environment related ezpublish configuration files.');
+        $output->writeln('Make sure to apply the config relevant to your install to you ezplatform.yml file.');
```
glye:
```suggestion
        $output->writeln('Make sure to apply the config relevant to your install to your ezplatform.yml file.');
```

bundle/Command/LegacyInitCommand.php
```diff
@@ -0,0 +1,250 @@
+<?php
+/**
+ * File containing the LegacySrcSymlinkCommand class.
```
glye:
```suggestion
 * File containing the LegacyInitCommand class.
```

bundle/Command/LegacyInitCommand.php
```diff
@@ -0,0 +1,250 @@
...
+use Symfony\Component\Console\Output\ConsoleOutputInterface;
+
+class LegacyInitCommand extends ContainerAwareCommand
+{
+    protected function configure()
+    {
+        $this
+            ->setName('ezpublish:legacy:init')
+            ->setDefinition(
+                array(
```
glye: Are we using old school array syntax here?

bundle/Command/LegacyInitCommand.php
```diff
@@ -0,0 +1,250 @@
...
+
+        /**
+         * @var \Symfony\Component\Filesystem\Filesystem
+         */
+        $filesystem = $this->getContainer()->get('filesystem');
+
+        $filesystem->mkdir([
+            $srcArg,
+            "$srcArg/design",
+            //"$srcArg/AppBundle",
```
glye: Testing code?

bundle/Command/LegacyInitCommand.php
```diff
@@ -0,0 +1,250 @@
...
+            "$srcArg/AppBundle/ezpublish_legacy",
+            "$srcArg/settings",
+            "$srcArg/settings/override",
+            "$srcArg/settings/siteaccess",
+        ]);
+
+        $output->writeln(<<<'EOT'
+
+The following folders should have been created (or where already present):
```
glye:
```suggestion
The following folders should have been created (or were already present):
```

INSTALL.md
```diff
@@ -3,76 +3,40 @@
 Instructions below will take you true installing legacy bridge and implicit legacy on top of a eZ Platform 1.13.x - 2.x
```
glye:
```suggestion
Instructions below will take you through installing legacy bridge and implicit legacy on top of a eZ Platform 1.13.x - 2.x
```
Off topic, but it's so grating :)

bundle/Command/LegacyInitCommand.php
andrerom: php-cs-fixer is not installed on this repo, it can be done and we can then cleanup CS across supported branches as followup to this.

bundle/Command/LegacyInitCommand.php
andrerom: Yes, kind of

andrerom: Fixed

emodric:
I'd like to see manual install instructions preserved in a separate file, as I don't like to rely on scripted installations 100%.

Otherwise, looks good.

andrerom:
@emodric Better?

PR #152, opened 2018-07-03
EZP-29144: Make websiteToolbarAction's Response cacheable

natanael89:
JIRA issue: [EZP-29144](https://jira.ez.no/browse/EZP-29144)

> Currently, the cache related to Website Toolbar doesn't take into the account if someone is an owner of the viewed Content Object. This means that if someone who doesn't have content/edit permissions access views this Content Object first, then the owner won't have an option to edit that Content Object using Website Toolbar.

The following PR: https://github.com/ezsystems/DemoBundle/pull/194 fixes the issue in the Symfony stack by loading the Website Toolbar using ESI, so that it won't be cached alongside the rest of the page content ("Vary: X-User-Hash" is used there, which is wrong when we need to deal with Owner ( Self ) Policy Limitation).

This PR makes websiteToolbarAction's Response cacheable with "Vary: Cookie". I used the content cache parameters to determine the headers.

### DIFF
--- a/bundle/Controller/WebsiteToolbarController.php
+++ b/bundle/Controller/WebsiteToolbarController.php
@@ -146,9 +146,7 @@ class WebsiteToolbarController extends Controller
             // reverse proxy caches the different possible variations of the
             // response as it can depend on user role for instance. X-User-Hash cannot
             // be used since the website toolbar can have Owner( Self ) Policy Limitation.
-            if ($request->headers->has('Cookie')) {
-                $response->setVary('Cookie');
-            }
+            $response->setVary('Cookie');
         }
 
         return $response;
--- a/bundle/Controller/WebsiteToolbarController.php
+++ b/bundle/Controller/WebsiteToolbarController.php
@@ -67,7 +67,7 @@ class WebsiteToolbarController extends Controller
      */
     public function websiteToolbarAction($locationId, Request $request)
     {
-        $response = new Response();
+        $response = $this->buildResponse();
 
         if (isset($this->csrfTokenManager)) {
             $parameters['form_token'] = $this->csrfTokenManager->getToken('legacy')->getValue();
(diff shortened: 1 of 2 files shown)


bundle/Controller/WebsiteToolbarController.php
```diff
@@ -109,6 +109,36 @@ public function websiteToolbarAction($locationId, Request $request)
...
+    /**
+     * Build the response so that depending on settings it's cacheable.
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    protected function buildResponse()
+    {
+        $request = $this->getRequest();
+        $response = new Response();
+        if ($this->getParameter('content.view_cache') === true) {
```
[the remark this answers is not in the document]
natanael89: Yes, that is better. Fixed, thanks.

bundle/Controller/WebsiteToolbarController.php
```diff
@@ -109,6 +124,36 @@ public function websiteToolbarAction($locationId, Request $request)
...
+                    $this->defaultTtl
+                );
+            }
+
+            // Make the response vary against Cookie header ensures that an HTTP
+            // reverse proxy caches the different possible variations of the
+            // response as it can depend on user role for instance. X-User-Hash cannot
+            // be used since the website toolbar can have Owner( Self ) Policy Limitation.
+            if ($request->headers->has('Cookie')) {
+                $response->setVary('Cookie');
```
andrerom: wondering if we should always set this even if there are no cookie, as otherwise the non varying cache will be set in varnish and used also by requests with cookies afaik.

natanael89: I'll check that to be sure and change it if needed.

andrerom: On further taught: In practice it provably won't happen _if_ eZ puts vary by user hash in request as well.

natanael89: @andrerom from my tests it looks like eZ indeed puts vary by user hash in the request. But otherwise, vary by cookie would be needed. So I'll change that fragment to always set it, just to be on the safe side.

bundle/Controller/WebsiteToolbarController.php
```diff
@@ -67,7 +82,7 @@ public function __construct(
      */
     public function websiteToolbarAction($locationId, Request $request)
     {
-        $response = new Response();
+        $response = $this->buildResponse();
```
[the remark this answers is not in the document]
natanael89: We don't need request in `buildResponse` method anymore, so I'm just going to remove that.

PR #153, opened 2018-07-08
Update tag for preview cache

zerustech:
The admin preview function does not display the draft content that is
currently being edited, this is because the preview cache is not purged,
and the preview function always fetches content from the view cache, 
refer to ezsystems/ezpublish-kernel#2396 for details.

To fix this issue, `PersistenceCachePurger.php::contentVersion()` has 
been changed to invalidate cache with the new tags.

| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | n/a
| **Bug**            | yes
| **New feature**    | no
| **Target version** | `v2.0.0`
| **BC breaks**      | no
| **Tests pass**     | yes
| **Doc needed**     | no

**TODO**:
- [ ] Implement feature / fix a bug.
- [ ] Implement tests.
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

### DIFF
--- a/bundle/Cache/PersistenceCachePurger.php
+++ b/bundle/Cache/PersistenceCachePurger.php
@@ -167,8 +167,9 @@ class PersistenceCachePurger implements CacheClearerInterface
             return;
         }
 
+        // todo Once we can require 2.2.2+, simplify to just invalidate "content-{$contentId}-version-{$versionNo}"
         $this->cache->deleteItem("ez-content-version-info-${contentId}-${versionNo}");
-        $this->cache->invalidateTags(["content-{$contentId}-version-{$versionNo}"]);
+        $this->cache->invalidateTags(["content-${contentId}-version-list", "content-{$contentId}-version-{$versionNo}"]);
     }
 
     /**
--- a/bundle/Cache/PersistenceCachePurger.php
+++ b/bundle/Cache/PersistenceCachePurger.php
@@ -168,7 +168,7 @@ class PersistenceCachePurger implements CacheClearerInterface
         }
 
         $this->cache->deleteItem("ez-content-version-info-${contentId}-${versionNo}");
-        $this->cache->invalidateTags(["content-${contentId}-version-list"]);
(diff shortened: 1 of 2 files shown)


andrerom:
Sorry for not giving input sooner, but your PR's made me aware we where lacking a way to invalidate versions properly.

So added that here: https://github.com/ezsystems/ezpublish-kernel/pull/2396

It means you can instead just edit exisitng `contentVersion()` and change it's last line to:
```php
         $this->cache->invalidateTags(["content-{$contentId}-version-{$versionNo}"]);
```


Which will now make sure to clear _all_ occurrences of the cache for the specific version.

Hope this makes sense to you to ;)

zerustech:
Hi @andrerom, I have revised the code accordingly. Thanks!

bundle/Cache/PersistenceCachePurger.php
```diff
@@ -168,7 +168,7 @@ public function contentVersion($contentId, $versionNo)
         }
 
         $this->cache->deleteItem("ez-content-version-info-${contentId}-${versionNo}");
```
andrerom: you can also drop this one

andrerom: Ignore that

andrerom: As we want to make sure this works also for <2.2.2, instead change the lines here to for instance:
```php
// todo Once we can require 2.2.2+, simplify to just invalidate "content-{$contentId}-version-{$versionNo}"
$this->cache->deleteItem("ez-content-version-info-${contentId}-${versionNo}");
$this->cache->invalidateTags(["content-${contentId}-version-list", "content-{$contentId}-version-{$versionNo}"]);
```

zerustech: @andrerom So I don't need do anything, do I?

andrerom: Not anymore at least ;)

PR #155, opened 2018-07-20
Throw exception when module or view is not found.

zerustech:
The legacy ezpublish does not return the 404 response code for any
request to a page that does not exist. It simply redirects all invalid
requests to the login page with a resonse code 200.

To fix this issue, method `ezpkernelweb::dispatchLoop()` has been
changed to throw exception when the requested module or view does not
exist. In order to redirect the invalid requests to a 404 error page,
the exception thrown by the legacy ezpublish needs to be converted to
`HttpNotFoundException`.

Refer to ezsystems/ezpublish-legacy#1379 for details.

| Question           | Answer
| ------------------ | ------------------
| **JIRA issue**     | n/a
| **Bug**            | yes
| **New feature**    | no
| **Target version** | `v2.0.0`
| **BC breaks**      | yes
| **Tests pass**     | yes
| **Doc needed**     | no

**TODO**:
- [ ] Implement feature / fix a bug.
- [ ] Implement tests.
- [ ] Fix new code according to Coding Standards (`$ composer fix-cs`).
- [x] Ask for Code Review.

### DIFF
--- a/bundle/EventListener/EzpModuleNotFoundToNotFoundListener.php
+++ b/bundle/EventListener/EzpModuleNotFoundToNotFoundListener.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * File containing the EzpModuleNotFoundToNotFoundListener class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace eZ\Bundle\EzPublishLegacyBundle\EventListener;
+
+use ezpModuleNotFound;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+class EzpModuleNotFoundToNotFoundListener implements EventSubscriberInterface
+{
+    public static function getSubscribedEvents()
+    {
+        return [
+            KernelEvents::EXCEPTION => ['onException', 512],
+        ];
+    }
+
+    public function onException(GetResponseForExceptionEvent $event)
+    {
+        if ($event->getException() instanceof ezpModuleNotFound) {
+            $event->setException(new NotFoundHttpException('Page Not Found', $event->getException()));
+        }
+    }
+}
--- a/bundle/EventListener/EzpModuleViewNotFoundToNotFoundListener.php
+++ b/bundle/EventListener/EzpModuleViewNotFoundToNotFoundListener.php
@@ -1,0 +1,32 @@
+<?php
+/**
+ * File containing the EzpModuleViewNotFoundToNotFoundListener class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+namespace eZ\Bundle\EzPublishLegacyBundle\EventListener;
+
+use ezpModuleViewNotFound;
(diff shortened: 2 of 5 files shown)


emodric:
Isn't converting legacy error pages to 404s already handled by legacy bridge?

andrerom:
I thought so, which is why I pinged you as I guess this is related to #127 _"[2.0]Remove deprecated features"_.

emodric:
Without going into why legacy kernel does what it does (returning a redirect instead of 404), I feel like this PR is not needed. `legacy_mode` `false` already handles conversion of 404 and unknown module view pages to Symfony HTTP exception, while `legacy_mode` `true` should retain any and all behaviour legacy kernel does, even if its redirecting instead of returning 404. That's what `legacy_mode` `true` is all about.

zerustech:
Hi @emodric, I didn't request the legacy kernel to return a 404 code, instead, I just suggested that it should throw the exception to trigger the twig exception controller "tjsoes.exception.controller:showAction". 

Currently, the legacy kernel simply redirects all invalid requests (unauthorized requests and authorized but invalid requests) to the login page, which is incorrect:
* http://<domain-name>/content/edit - redirects to the login page (this makes sense because the user is not allowed to access content/edit);

* http://<domain-name>/abc - redirects to the login page, which is incorrect because module "abc" does not exist, and the web developer should have an opportunity the report an 404 error;

* http://<domain-name>/content/abc - redirects to the login page, which is also incorrect because the view "content/abc" does not exist.

emodric:
Okay, but what I'm saying here is that `legacy_mode` `true` should do whatever legacy kernel does, without any exceptions, that's the point of `legacy_mode` being `true`. So any invalid behaviour should be fixed in the legacy kernel itself. Once it's fixed, if you need a Symfony 404, you should switch your siteaccess to `legacy_mode` `false`, which will then convert any "Module not found" or "View not found" errors into Symony 404s. When legacy mode is *enabled*, conversion should not happen.

zerustech:
@emodric 
> ... convert any "Module not found" or "View not found" errors into Symony 404s. 

That's exactly what I'm proposing in this PR. The "Module not found" (`ezpModuleNotFound`) or "View not found" (`ezpModuleViewNotFound`) errors are converted into Symfony 404 errors (`HttpNotFoundException`). The `ezpModuleNotFound` and `ezpModuleViewNotFound`
 exceptions are thrown by the ezpublish legacy in ezsystems/ezpublish-legacy#1379

unless you think it's better to let ezpublish legacy directly throw `HttpNotFoundException`. If this is the case, then there is no need to change ezsystems/legacy-bridge. Please let me know what do you think. Thanks

emodric:
This PR does not take into account `legacy_mode` configuration. Conversion cannot happen when legacy mode is enabled, to preserve the behaviour of legacy kernel.

When/If (see my comment over there) legacy kernel PR to throw `ezpModuleNotFound` and `ezpViewNotFound` is accepted, this one will need to be modified to do the conversion only if legacy mode is disabled.

Additionally, if [this code](https://github.com/ezsystems/LegacyBridge/blob/master/bundle/LegacyResponse/LegacyResponseManager.php#L99-L104) already does the conversion, then I think this PR is not needed at all, regardless of what legacy kernel does.

zerustech:
@emodric, Thanks for your feedback. I didn't know the `LegacyResponseManager` generates `NotFoundHttpException` based on the `errorCode` in `$moduleResult`. If this is the case, then it is possible to only revise the `ezpublish-legacy` and close this PR. I will review the code of `ezpkernelweb.php` and get back to you soon.

zerustech:
@emodric I made a mistake, it turned out to be a configuration issue, so I'm closing this PR. Thanks a lot for your support.

The cause of this problem is as follows:
The front end is handled by Symfony, and a lot of ini setting entries have been overridden by the Symfony configuration files as well, so I was trying to cleanup the ini settings for the siteaccess-specific ini files and I had unfortunately **"overly optimized"** the site.ini and the `RequireUserLogin=false` entry was removed incorrectly.

PR #157, opened 2018-08-10
Fix SiteAccess mapping on inline legacy calls with default site access

andrerom:
_Also enable travis for 1.4 and fix regressions in unit tests since it last ran last fall._

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ cache:
 branches:
   only:
     - master
+    - 1.4
 
 # setup requirements for running unit tests
 before_script:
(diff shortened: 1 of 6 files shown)


emodric:
What does inline legacy call mean? :)

andrerom:
Ah, as in inline legacy fallbacks, so when legacy mode is disabled but you rely on fallback to legacy for some content views, full or (in-) line as part of full ContentView on platform.

In this case issue was links generated by legacy on template output for ezurl would have `/_fragment` as prefix.

PR #159, opened 2018-09-03
Check if legacy engine is set before calling it

emodric:
On PHP 7.3, `loadTemplate` method is somehow called before `setEzLegacyEngine` had a chance to set the legacy engine to the `Environment` class. Checking for `null` on `$this->legacyEngine` does the job apparently, with no side-effects.


emodric:
I tried to debug, but didn't have much time to go into details.

PR #163, opened 2019-01-15
Fix CORS OPTIONS Call for POST REST requests

SalvatorePollaci:
I have found that all OPTIONS calls for POST REST requests return 405 error message, this is due to the fact that DefaultRouter uses Symfony's default RequestContext which refers to a GET request. Therefore I'm setting the the RequestContext given the Request Object.


andrerom:
+1 should this be applied to 1,4 as well? _(for eZ Platform 1.x)_

andrerom:
@SalvatorePollaci Up for rebasing, or do you prefer we do that for the merge? :)

andrerom:
ok

andrerom:
Fixed in https://github.com/ezsystems/LegacyBridge/commit/ab86627db35e93309572b3fac70fd231ebbc86dd

PR #170, opened 2019-05-02
[Feature] Inject Debug settings

andrerom:


bundle/LegacyMapper/Configuration.php
```diff
@@ -340,4 +343,36 @@ private function getClusterSettings()
 
         return $clusterSettings;
     }
+
+    private function getDebugSettings()
+    {
+        // if disabled, we don't touche any settings
+        if (!$this->container->getParameter('kernel.debug')) {
```
emodric: I'd like to see a special parameter controlling this, nut just `kernel.debug`. In our case, we don't develop on legacy any more, so activating debug in legacy purely because Sf debug is active is counterproductive and slows down operation.

andrerom: > I'd like to see a special parameter controlling this
I'm not going to add that, then people can just as well edit back and forward their site.ini override so they don't need to learn a new setting for this.

What I aim for here is something minimal default debug output and DevelopmentMode settings that makes sense to enable when symfony is also in debug. Are you saying you are against all of thee or do you see some could be ok and others not?

emodric: Any one of these. Each of them bring their own performance penalty (small or large) for debug features we don't need.

andrerom: well that is kernel.debug is exactly what kernel.debug and related setting does as well, but ok

emodric: Sure, but as I said, Sf debug we need in dev, eZ legacy we don't :)

PR #171, opened 2019-05-03
 [Optimization] Skip loading user and path if not needed

andrerom:
Some optimizations to reduce default Repo lookups by bridge from 11 to 3 cache lookups , this also avoids the serialization and mapping of these objects to API objects and logic around that.


andrerom:
@emodric Can you review this despite the Travis PHP 5.6 proc_open() failure?

PR #172, opened 2019-05-03
[Feature] Add option to Init to inject basic INI settings

andrerom:
Basic settings to simplify testing legacy / bridge on a eZ Platform install.
Config here aligns with what is in clean eZ Platform install, and adds a `legacy_admin` siteaccess to access legacy admin.

The generated ini contains debug and development settings so it's easy to switch them on if needed.


bundle/Resources/init_ini/siteaccess/legacy_admin/site.ini.append.php
```diff
@@ -0,0 +1,33 @@
...
+
+[SiteSettings]
+DefaultPage=content/dashboard
+LoginPage=custom
+
+[SiteAccessSettings]
+RequireUserLogin=true
+RelatedSiteAccessList[]
+RelatedSiteAccessList[]=site
+#RelatedSiteAccessList[]=admin
```
emodric: Why commented out?

bundle/Resources/init_ini/siteaccess/legacy_admin/site.ini.append.php
```diff
@@ -0,0 +1,33 @@
...
+
+[DesignSettings]
+SiteDesign=admin2
+AdditionalSiteDesignList[]
+AdditionalSiteDesignList[]=admin
+AdditionalSiteDesignList[]=standard
+AdditionalSiteDesignList[]=base
+
+[RegionalSettings]
+Locale=eng-GB
```
emodric: You could in theory make this a Twig template and inject the correct values taken from Symfony container. Same goes for the list of siteaccesses.

Similar to this: https://github.com/netgen/NetgenAdminUIBundle/blob/master/bundle/Installer/_templates/legacy_siteaccess/site.ini.append.php#L23

bundle/Resources/init_ini/siteaccess/legacy_admin/site.ini.append.php
andrerom: admin is not a legacy siteaccess, so I was unsure if it should be here or not. Probably not.

bundle/Resources/init_ini/siteaccess/legacy_admin/site.ini.append.php
andrerom: Ideally yes.

Is that something you1'd be able to contribute here? _(the code involved in the "rendering" and saving of the results of the ini templates)_

emodric: I could, but next only next week :+1:

bundle/Resources/init_ini/siteaccess/legacy_admin/site.ini.append.php
emodric: We always kept it, didn't matter. Probably doesn't matter either way.

andrerom: well, it's used misc places to provide list of siteaccess to see things, like preview, template view and ini view. Which is why I think removing it might be best.

andrerom: Removed it given it was not present in override settings.

bundle/Resources/init_ini/siteaccess/legacy_admin/site.ini.append.php
andrerom: @emodric can we do this as a followup? Given this is meant to be used for testing use cases straight after clean install, and atm clean install still is limited to eng-GB, this is imho ok for a start.

emodric: Sure, no problem.

andrerom: @emodric I'm unable to figure out the proc_open issue on PHP 5.6 omn Travis, could you review this regardless of 5.6 failure?

PR #175, opened 2019-12-03
Remove visibility change on global helper core service override.

xserna:
PR coming from https://github.com/ezsystems/LegacyBridge/pull/126#discussion_r353142084

### DIFF
--- a/bundle/Resources/config/templating.yml
+++ b/bundle/Resources/config/templating.yml
@@ -66,6 +66,8 @@ services:
     # Overriding core helper
     ezpublish.templating.global_helper:
         alias: ezpublish.templating.global_helper.legacy
+        # Needs to be public to use the service as a Twig global variable
+        public: true
 
     templating.engine.eztpl:
         class: eZ\Publish\Core\MVC\Legacy\Templating\LegacyEngine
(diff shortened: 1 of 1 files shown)


emodric:
Can you use `public: true` to mark it specifically as public, instead of relying on the default Symfony behaviour (and to avoid the deprecation message)?

You can also leave a comment why it is needed to be public.

andrerom:
@emodric Maybe you can use suggest feature for what you had in mind?

bundle/Resources/config/templating.yml
```diff
@@ -66,7 +66,6 @@ services:
     # Overriding core helper
     ezpublish.templating.global_helper:
         alias: ezpublish.templating.global_helper.legacy
```
emodric:
```suggestion
        alias: ezpublish.templating.global_helper.legacy
        # Needs to be public to use the service as a Twig global variable
        public: true
```

emodric: @andrerom Here you go. Care should be taken to not end up with duplicate `public` option, as I couldn't add the suggestion to a deleted line.

andrerom: @xserna Could you update PR with suggested change above?

xserna: > @xserna Could you update PR with suggested change above?

Sure, I've this on my todo list. Sorry for the delay, I'll update it now.

andrerom:
Thanks @xserna 👏

PR #180, opened 2020-11-10
Update php-cs-fixer (2.14.2 is not compatible with Composer 2)

blankse:
php-cs-fixer has following error with Composer 2:
`Undefined index: name`

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/4905


andrerom:
Thanks @blankse !

PR #184, opened 2021-07-27
IBX-819: Shortened persistence cache tags to save memory usage

konradoboza:
JIRA ticket: [IBX-819](https://issues.ibexa.co/browse/IBX-819).

Depends on: https://github.com/ezsystems/ezpublish-kernel/pull/3114


emodric:
This might be seen as a breaking change. Especially, since this is a legacy related project, not developed actively any more and in bugfix mode only, I presume.

konradoboza:
Shortening tags aims to reduce RAM usage. If we merge https://github.com/ezsystems/ezpublish-kernel/pull/3114 we also need to align persistence cache purging done within Legacy Bridge.

alongosz (approved):
There's no BC promise on cache tag names, they're kind of "runtime" information. You need to purge your cached information when upgrading your project. Otherwise it might lead to issues which are difficult to predict and debug. That was always the case, not just for this change.

Steveb-p (approved):

micszo (approved):
Tested with https://github.com/ezsystems/ezpublish-kernel/pull/3114.

